PR #223, opened 2014-12-17
[WIP] Spreadsheet validation

robbiejackson:
This is associated with moving inappropriate code out of config/settings.yml which is associated with the upload of orphan spreadsheets (https://osraav.atlassian.net/browse/OSRA-240)
I've created this pull request just to provide visibility of the approach I've been taking, to see if you want me to progress along these lines. 
Note that this will cause several rspec failures, because I haven't updated the tests to align with this module design.


spec/lib/excel_upload_spec.rb
```diff
@@ -3,28 +3,28 @@
 
 describe ExcelUpload do
   before :each do
-    @settings = Settings.import
+    @first_row = 4
```
robbiejackson: woops - I should be getting the class variable here instead of assigning @first_row to 4 directly.

lib/orphan_importer.rb
```diff
@@ -1,52 +1,154 @@
...
+#   - an array of the import errors identified
+#  
+  @@yn_to_bool = { 'y' => true, 'Y' => true, 'n' => false, 'N' => false }
+  @@male_arabic_yn_to_bool = { 'موجود' => true, 'متوفى' => false, 'y' => true, 'Y' => true, 'n' => false, 'N' => false }
+  @@female_arabic_yn_to_bool = { 'موجودة' => true, 'متوفاة' => false, 'y' => true, 'Y' => true, 'n' => false, 'N' => false }
+  @@gender = { 'ذكر' => 'Male', 'أنثى' => 'Female'}
+  @@province_to_code = { 'دمشق وريف دمشق' => 11, 'حلب' => 12, 'حمص' => 13, 'حماة' => 14, 'اللاذقية' => 15,
+    'دير الزور' => 16, 'درعا' => 17, 'إدلب' => 18, 'الرقة' => 19, 'الحسكة' => 20, 'طرطوس' => 21,
+    'السويداء' => 22, 'القنيطرة' => 23, 'خارج سوريا' => 29 } 
+  @@mapper =
```
PurityControl: One of the goals here was to lose the fixed spreadsheet column position (eg 'A', 'B', 'C') and replace it with a columns title that would accept both Arabic and English. Could the format be modified to achieve this ?

robbiejackson: Yes I've checked and it would be easy to change to drive the mapping off the column titles. (My only doubt was how it handled the merged cells.)
Nik, could you confirm with the clients that they'd want to go down this route? It would mean that they wouldn't be able to change the header titles in the first 3 rows without impacting the software. It might also then be better to have this mapping in a config file, rather than in a ruby file. Is there a natural place to put such a file within the hierarchy? What do you guys think?

robbiejackson:
Functionality has been updated to drive the mapping from spreadsheet to orphan attributes off the headers within the spreadsheet. Also a number of test spreadsheets updated to take account of the latest changes within the v3.3 headers.

PR #250, opened 2014-12-31
Osra290 create partner show

theB264:
Added a partner#show controller and a view, added routing specs and namespaced the existing ones, routed the orphan_lists#index nested under partner#show


spec/controllers/hq/partners_controller_spec.rb
```diff
@@ -16,4 +16,29 @@ def foobar_action
...
+      sign_in instance_double(AdminUser)
+    end
+
+    describe 'GET #show' do
+      before :each do
+        FactoryGirl.create(:partner)
+      end
+
+      example 'invalid :id should redirect & set flash' do
+        get :show, id: (Partner.all.count + 1)
```
PurityControl: agreed in addition id's in database aren't guaranteed to be less than the record count. There are better ways to setup Partner not being found.

spec/controllers/hq/partners_controller_spec.rb
```diff
@@ -16,4 +16,29 @@ def foobar_action
...
+        expect(response).to redirect_to hq_partners_path
+        expect(flash[:error]).to match /Cannot find Partner/
+      end
+
+      example 'valid :id should not redirect nor set flash' do
+        get :show, id: Partner.first.id
+        expect(response.status).to_not be 302
+        expect(flash[:error]).to_not match /Cannot find Partner/
+      end
+    end
```
PurityControl: agreed in addition the presence of a passing happy path makes tthe check in this test unnecessary

spec/controllers/hq/partners_controller_spec.rb
```diff
@@ -16,4 +16,32 @@ def foobar_action
...
   end
+
+  describe 'while signed-in' do
+    before :each do
+      sign_in instance_double(AdminUser)
+    end
+
+    describe 'GET #show' do
+      before :each do
+        @id= FactoryGirl.create(:partner).id
```
PurityControl: should this not be in a let?

app/controllers/hq/partners_controller.rb
```diff
@@ -5,4 +5,11 @@ class Hq::PartnersController < ApplicationController
   def index
     @partners= Partner.all
   end
+
+  def show
+    if not (@partner= Partner.find_by_id(params[:id]))
```
PurityControl: still tempted to use find and let rails deal with the rare case of an incorrect id being given

theB264: `Model.find` will raise an exception on invalid id, is that which method you refer to?

PurityControl: yup

NikitaAvvakumov: Is this a good time to think about Null Objects?

Otherwise, I agree with @PurityControl: while this failure is somewhat more graceful than a simple app breakdown, it gives the user no actionable information.

Also, are we now really opposed to `unless`?

app/views/layouts/_footer.html.erb
```diff
@@ -1,4 +1,3 @@
-
-    <div class="footer">
-      <p>Powered by <a href="http://www.agileventures.org">AgileVentures</a> volunteers &#9734; Open Source software</p>
-    </div>
+<div class="footer">
+  <p>Powered by <a href="http://www.agileventures.org">AgileVentures</a> volunteers &#9734; Open Source software</p>
+</div>
```
NikitaAvvakumov: Given that this is not a public-facing site, could the footer provide different info (e.g. link to the G+ support community) or be dispensed with altogether?

app/views/hq/partners/index.html.erb
```diff
@@ -35,7 +35,7 @@
                 <%= partner.status.name %>
               </td>
               <td>
-                <%= partner.start_date.strftime('%B %d, %Y') %>
+                <%= partner.start_date.strftime('%d %B %Y') %>
```
NikitaAvvakumov: Time to extract this formatting into a helper?

app/views/hq/partners/show.html.erb
```diff
@@ -0,0 +1,44 @@
...
+  <dt>OSRA Num</dt>
+  <dd><%= @partner.osra_num %></dd>
+</dl>
+<dl class="dl-horizontal">
+  <dt>Status</dt>
+  <dd><%= @partner.status.name %></dd>
+</dl>
+<dl class="dl-horizontal">
+  <dt>Start Date</dt>
+  <dd><%= @partner.start_date.strftime('%d %B %Y') %></dd>
```
NikitaAvvakumov: As above

spec/controllers/hq/partners_controller_spec.rb
NikitaAvvakumov: Prefer to move it into the `with valid :id` block

spec/views/hq/partners/show_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe "hq/partners/show.html.erb", type: :view do
+  let(:partner) do
+    FactoryGirl.build_stubbed :partner, created_at: Date.current, updated_at: Date.current
+  end
+
+  context 'orphan lists link' do
```
NikitaAvvakumov: Prefer `describe` here

spec/views/hq/partners/show_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe "hq/partners/show.html.erb", type: :view do
+  let(:partner) do
+    FactoryGirl.build_stubbed :partner, created_at: Date.current, updated_at: Date.current
```
NikitaAvvakumov: Why is it necessary to specify the dates?

spec/routing/hq/partners_routing_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+
+RSpec.describe 'Hq::Partners routing', type: :routing do
+  specify { expect(get: '/hq/partners').          to be_routable }
+  specify { expect(post: '/hq/partners').         to be_routable }
+  specify { expect(get: '/hq/partners/new').      to be_routable }
+  specify { expect(get: '/hq/partners/1').        to be_routable }
+  specify { expect(get: '/hq/partners/1/edit').   to be_routable }
+  specify { expect(put: '/hq/partners/1').        to be_routable }
+  specify { expect(patch: '/hq/partners/1').      to be_routable }
+  specify { expect(delete: '/hq/partners/1'). not_to be_routable }
```
NikitaAvvakumov: Would anyone else find it useful to add comments like `#index`, `#edit`, `#destroy` etc. to each line?

app/views/hq/partners/show.html.erb
```diff
@@ -0,0 +1,44 @@
+<dl class="dl-horizontal">
```
NikitaAvvakumov: For all PRs that implement or change views, screenshots would be great

spec/controllers/hq/partners_controller_spec.rb
NikitaAvvakumov: `nil` comes to mind

spec/controllers/hq/partners_controller_spec.rb
```diff
@@ -16,4 +16,32 @@ def foobar_action
...
+      end
+
+      describe 'redirect & flash' do
+        example 'with invalid :id' do
+          get :show, id: 0
```
NikitaAvvakumov: Would `nil` be more intention-revealing?

spec/routing/hq/partners_routing_spec.rb
PurityControl: I think its a good idea. I wouldn't miss it if it wasn't there but these little things can stop you losing your cadence when in your workflow.

razorcd: I am pro these comments. And in general we could use more code comments in osra.

app/views/hq/partners/index.html.erb
azangru: We even have a helper. It's in https://github.com/AgileVentures/osra/blob/develop/app/helpers/admin/time_helper.rb

app/controllers/hq/partners_controller.rb
theB264: Re: `unless`
I avoided using it here because some members of the team dislike it.  Since I wasn't using it for a return value (the only time it's different than `if not`) I just omitted it

app/views/layouts/_footer.html.erb
theB264: I merely fixed the white space here, the footer contents are beyond the scope of this PR.  Shall we create a task?

app/views/hq/partners/index.html.erb
theB264: OSRA318

app/views/hq/partners/show.html.erb
theB264: OSRA318

spec/views/hq/partners/show_spec.rb
theB264: `build_stubbed` inserts nil dates while `create` uses real dates

spec/controllers/hq/partners_controller_spec.rb
theB264: I tried `nil` first, it disallows that

app/views/hq/partners/index.html.erb
theB264: Also this helper shouldn't die if date is nil, and probably shouldn't be namespaced under Admin

spec/controllers/hq/partners_controller_spec.rb
razorcd: i think it only accepts values that can be transformed in strings, because queries are strings.
and nil.to_s returns empty string

theB264: Then perhaps `''` would have been better to use.  These tests are gone now anyway due to code review.

app/controllers/hq/partners_controller.rb
razorcd: @PurityControl created some nice methods in the controller, like `load_partner`. Would it be better to use those here too?
https://github.com/AgileVentures/osra/pull/252/files

theB264: These don't exist yet.  His pr is based on this one

spec/controllers/hq/partners_controller_spec.rb
razorcd: using `''` will give a path without an id so `params[:id]` will not exist.  (/partners/show/)
But if the tests are go then it's gooone! :)

theB264:
screenshot from 2015-01-02 13 14 44

app/views/layouts/_footer.html.erb
NikitaAvvakumov: If the team thinks it's a good idea. I didn't mean for it to be a part of this PR, just wrote down my thoughts - should have made that clear.

NikitaAvvakumov:
Odd, I expected `dl-horizontal` to look different - any idea why it's wrapping?
Anyway, the PR looks good to go.

theB264:
Not sure why it's wrapping, but it doesn't have CSS or eye-candy yet either

PR #251, opened 2014-12-31
partial for partners

PurityControl:
In the interests of keeping pr's small here is one just for the partial to allow the edit/update and new/create stories to run in parallel.

- error messages in form to be implemented in a later pr abstracted out into
  custom form helper to be used across application
- have had to require CGI in tests to escape provinces for tests - is there a
  better way
- have used minitest assert_select in tests because that is my preference but
  I am open to suggestions if this offends people

See screenshot below:-

form partial 2014-12-31 07 56 19

### DIFF
--- a/spec/views/hq/partners/_form.html.erb_spec.rb
+++ b/spec/views/hq/partners/_form.html.erb_spec.rb
@@ -1,0 +1,73 @@
+require 'rails_helper'
+require 'cgi'
+
+RSpec.describe "hq/partners/_form.html.erb", type: :view do
+
+  let(:provinces) { Province.all }
+  let(:statuses) { Status.all }
+  let(:partner) { build_stubbed :partner }
+
+  before :each do
+    assign(:provinces, provinces)
+    assign(:statuses, statuses)
+    assign(:partner, partner)
+  end
+
+  specify 'has a form' do
+    render
+
+    assert_select "form"
+  end
+
+  context 'render value' do
+    specify 'Name' do
+      render
+
+      assert_select "input#partner_name", value: partner.name
+    end
+
+    specify 'Region' do
+      render
+
+      assert_select "input#partner_region", value: partner.region
+    end
+
+    specify 'Contact Details' do
+      render
+
+      assert_select "input#partner_name", value: partner.contact_details
+    end
+
+    specify 'Province' do
+      render
+
+      assert_select "select#partner_province_id" do
+        assert_select "option", value: provinces.first.id,
+                                 html: CGI::escape_html(provinces.first.name)
+      end
+    end
+
+    specify 'Status' do
+      render
+
+      assert_select "select#partner_status_id" do
+        assert_select "option", value: statuses.first.id,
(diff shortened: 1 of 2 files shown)


spec/views/hq/partners/_form.html.erb_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+require 'cgi'
+
+RSpec.describe "hq/partners/_form.html.erb", type: :view do
+  before :each do
+    @provinces = Province.all
+    @statuses = Status.all
+    @partner = build_stubbed :partner
+  end
+
+  context 'render the form and show data in form controls' do
```
theB264: `context 'renders value' do
  example 'Name' do`
This seems more clear in the code, test descriptions, and object heirarchy

theB264:
:+1:

NikitaAvvakumov:
Very happy with the small focused PR, and the partial looks good.
What's our final decision on view specs? At one point, we were leaning towards limiting them to branches & conditionals - i.e. testing any _logic_ that might be in a view but nothing else.

PurityControl:
I think possibly the only test that merits considering dropping is the first test (the form check), but I think it still has value. Given that its a partial used in multiple places I think checking input fields are present and render the appropriate values is prudent.
There shouldn't be much overhead given the exclusive use of fixture data and build_stubbed calls.
In #252 the only test I'm running is that the partial is called, because all the testing is effectively in the partial.
Given we should be ruthlessly minimising any logic in views I think we do need some (minimal) content checking, especially as cucumber should not be doing this. However, it needs to be abstract enough not to require rewriting for every change but give enough confidence that the right information in the view is being rendered.
If you think I have moved away from the vision of how our view test should be, do you want to put something on the table - a list say of what you envisioned seeing, and then we can move closer to it.

spec/views/hq/partners/_form.html.erb_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    specify 'Region' do
+      render
+
+      assert_select "input#partner_region", value: partner.region
+    end
+
+    specify 'Contact Details' do
+      render
+
+      assert_select "input#partner_name", value: partner.contact_details
```
theB264: typo

PurityControl:
The asserts have to be multiline, @NikitaAvvakumov typo catch made me spot a floor in the tests. However I have refactored out the render into the before block and reduced them to a single test to make things more readable and compact.

spec/views/hq/partners/_form.html.erb_spec.rb
```diff
@@ -0,0 +1,53 @@
+require 'rails_helper'
+require 'cgi'
+
+RSpec.describe "hq/partners/_form.html.erb", type: :view do
+
+  let(:provinces) { Province.all }
+  let(:statuses) { Status.all }
```
NikitaAvvakumov: Is there a performance hit from using `.all` rather than taking just one or two? I suspect not.

PR #252, opened 2015-01-01
Edit and Update partners https://osraav.atlassian.net/browse/OSRA-292

PurityControl:
implement partner edit update  https://osraav.atlassian.net/browse/OSRA-292

- update redirects to show view so this is dependent on OSRA-290 #250 
- partial for partners needs merged first #251 
- bootstrap-date-picker gem extracted out into its own pr, merge #259 first
Screenshot below:-

screen shot 2015-01-01 at 18 38 53

### DIFF
--- a/app/controllers/hq/partners_controller.rb
+++ b/app/controllers/hq/partners_controller.rb
@@ -17,13 +17,14 @@ class Hq::PartnersController < ApplicationController
 
   def update
     load_partner
+    load_associations
     build_partner
     save_partner or render 'edit'
   end
 
-  private
+private
   def load_partner
-    @partner ||= Partner.find(params[:id])
+    @partner = Partner.find(params[:id])
   end
 
   def load_associations
@@ -38,6 +39,7 @@ class Hq::PartnersController < ApplicationController
 
   def save_partner
     if @partner.save
+      flash[:success] = 'Partner successfuly saved'
       redirect_to hq_partner_url(@partner)
     end
   end
--- a/app/controllers/hq/partners_controller.rb
+++ b/app/controllers/hq/partners_controller.rb
@@ -9,4 +9,47 @@ class Hq::PartnersController < ApplicationController
   def show
     @partner= Partner.find(params[:id])
   end
+
+  def edit
+    load_partner
+    load_associations
+  end
+
+  def update
+    load_partner
+    build_partner
+    save_partner or render 'edit'
+  end
+
+  private
+  def load_partner
+    @partner ||= Partner.find(params[:id])
+  end
+
+  def load_associations
+    @provinces = Province.all
+    @statuses = Status.all
+  end
+
+  def build_partner
+    @partner ||= Partner.build
+    @partner.attributes = partner_params
+  end
+
+  def save_partner
+    if @partner.save
+      redirect_to hq_partner_url(@partner)
+    end
+  end
+
+  def partner_params
+    partner_params = params[:partner]
+    if partner_params
+      partner_params.permit(:name, :region, :contact_details, :province_id,
+                            :status_id, :start_date)
+    else
+      {}
+    end
+  end
+
 end
--- a/app/views/hq/partners/edit.html.erb
+++ b/app/views/hq/partners/edit.html.erb
@@ -1,0 +1,2 @@
+<h1> Edit Partner with Osra Number:- <%= @partner.osra_num %> </h1>
+<%= render partial: 'form' %>
--- a/app/views/hq/partners/edit.html.erb
+++ b/app/views/hq/partners/edit.html.erb
@@ -1,2 +1,2 @@
-<h1> Edit Partner with Osra Number:- <%= @partner.osra_num %> </h1>
+<h1> Edit Partner <%= @partner.name %>, OSRA number <%= @partner.osra_num %> </h1>
 <%= render partial: 'form' %>
--- a/spec/controllers/hq/partners_controller_spec.rb
+++ b/spec/controllers/hq/partners_controller_spec.rb
@@ -18,3 +18,28 @@ RSpec.describe Hq::PartnersController, type: :controller do
   end
 
 end
+
+RSpec.describe Hq::PartnersController, type: :controller do
+  before :each do
+    sign_in instance_double(AdminUser)
+    @partner = build_stubbed :partner
+    expect(Partner).to receive(:find).and_return(@partner)
+  end
+
+  it 'editing renders the edit view' do
+    get :edit, id: @partner.id
+    expect(response).to render_template 'edit'
+  end
+
+  it 'unsuccessful update renders the edit view' do
+    expect(@partner).to receive(:save).and_return(false)
+    patch :update, id: @partner.id, partner: {name: ""}
+    expect(response).to render_template 'edit'
+  end
+
+  it 'successful update redirects to the show view' do
+    expect(@partner).to receive(:save).and_return(true)
+    patch :update, id: @partner.id
+    expect(response).to redirect_to(hq_partner_path(@partner))
+  end
+end
--- a/spec/controllers/hq/partners_controller_spec.rb
+++ b/spec/controllers/hq/partners_controller_spec.rb
@@ -26,18 +26,18 @@ RSpec.describe Hq::PartnersController, type: :controller do
     expect(Partner).to receive(:find).and_return(@partner)
   end
 
-  it 'editing renders the edit view' do
+  specify 'editing renders the edit view' do
     get :edit, id: @partner.id
     expect(response).to render_template 'edit'
   end
 
-  it 'unsuccessful update renders the edit view' do
+  specify 'unsuccessful update renders the edit view' do
     expect(@partner).to receive(:save).and_return(false)
-    patch :update, id: @partner.id, partner: {name: ""}
+    patch :update, id: @partner.id
     expect(response).to render_template 'edit'
   end
 
-  it 'successful update redirects to the show view' do
+  specify 'successful update redirects to the show view' do
     expect(@partner).to receive(:save).and_return(true)
     patch :update, id: @partner.id
     expect(response).to redirect_to(hq_partner_path(@partner))
--- a/app/assets/javascripts/date-picker.js.coffee
+++ b/app/assets/javascripts/date-picker.js.coffee
@@ -1,0 +1,5 @@
+jQuery ->
+  $(".date-picker").datepicker({
+    format: "yyyy-mm-dd"
+    autoclose: true
+  })
--- a/app/views/hq/partners/_form.html.erb
+++ b/app/views/hq/partners/_form.html.erb
@@ -23,7 +23,7 @@
   </div>
   <div class="form-group">
     <%= f.label :start_date %>
-    <%= f.text_field :start_date, class: "form-control" %>
+    <%= f.text_field :start_date, class: "form-control date-picker" %>
   </div>
   <div class="form-group">
     <%= f.submit class: "btn btn-default" %>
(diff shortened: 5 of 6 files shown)


app/controllers/hq/partners_controller.rb
```diff
@@ -5,4 +5,47 @@ class Hq::PartnersController < ApplicationController
...
+  def update
+    load_partner
+    build_partner
+    save_partner or render 'edit'
+  end
+
+  private
+  def load_partner
+    @partner ||= Partner.find(params[:id])
+  end
```
razorcd: this method returns a partner or throws an errors. We should return partner or nil.

razorcd:
There are no flash messages for informing the user about successful or failed operations.

app/controllers/hq/partners_controller.rb
PurityControl: which method?

PurityControl:
The plan for handling failures is discussed in #251

app/controllers/hq/partners_controller.rb
razorcd: method  load_partner
Partner.find(9999) will throw an error

theB264: `.find_by_id` I believe will return an instance of Partner or `nil`

PurityControl: I'm not sure we do wan't to return nil, but that is probably a discussion better for a scrum

PurityControl: screen shot 2015-01-01 at 19 49 03

razorcd: I was thinking:

def load_partner
  @partner ||= Partner.find_by_id(params[:id])
end

to return a partner or nil and then use it like you did for save:

load_partner or redirect_to(partners_path  , notice: "The specified partner was not found")

This way it does not leave the app. An it is useful for any future situations.

PurityControl:
addressed discussions in scrum and pushed up date picker which requires another gem.
This would be a good time to discuss whether we have controller specific javascript with
 `<%= javascript_include_tag params[:controller] %> `

screen shot 2015-01-02 at 00 10 59

Gemfile
```diff
@@ -8,6 +8,8 @@ gem 'pg'
 gem 'sass-rails', '~> 5.0.0'
 # sass powered version of bootstrap https://github.com/twbs/bootstrap-sass
 gem 'bootstrap-sass', '~> 3.3.1'
+# jquery datepicker and bootstrap-sass gem don't play nicely
```
PurityControl: bootstrap-datepicker-rails stuff could be pulled out into its own pr

app/controllers/hq/partners_controller.rb
```diff
@@ -5,4 +5,48 @@ class Hq::PartnersController < ApplicationController
...
   end
+
+  def edit
+    load_partner
+    load_associations
+  end
+
+  def update
+    load_partner
+    load_associations
```
NikitaAvvakumov: Do you need to `load_associations` here for `render 'edit'`?

Do we want to try following Sandi Metz's rule that a controller should not instantiate more than one object? And adopt e.g. [this solution](http://robots.thoughtbot.com/sandi-metz-rules-for-developers) (bottom of page).

app/controllers/hq/partners_controller.rb
```diff
@@ -5,4 +5,48 @@ class Hq::PartnersController < ApplicationController
...
+    build_partner
+    save_partner or render 'edit'
+  end
+
+private
+  def load_partner
```
NikitaAvvakumov: Just semantics, but what about `assign` in favour of `load`? It seems more intention revealing.

app/controllers/hq/partners_controller.rb
```diff
@@ -5,4 +5,48 @@ class Hq::PartnersController < ApplicationController
...
+  end
+
+  def build_partner
+    @partner ||= Partner.build
+    @partner.attributes = partner_params
+  end
+
+  def save_partner
+    if @partner.save
+      redirect_to hq_partner_url(@partner)
```
NikitaAvvakumov: Flash?

app/controllers/hq/partners_controller.rb
```diff
@@ -5,4 +5,48 @@ class Hq::PartnersController < ApplicationController
...
+    end
+  end
+
+  def partner_params
+    partner_params = params[:partner]
+    if partner_params
+      partner_params.permit(:name, :region, :contact_details, :province_id,
+                            :status_id, :start_date)
+    else
+      {}
```
NikitaAvvakumov: Why would we want the controller to carry on with an empty params hash when one should be present? i.e. Why this instead of `params.require(:partner).permit(...)`?

app/controllers/hq/partners_controller.rb
PurityControl: yes I need it in case an edit validation fails.
So to achieve this goal are you talking about creating a new class that references the three objects as part of that single object instantiation?

app/views/hq/partners/edit.html.erb
```diff
@@ -0,0 +1,2 @@
+<h1> Edit Partner with Osra Number:- <%= @partner.osra_num %> </h1>
```
NikitaAvvakumov: ":-" is uncommon - something special? I can see there being no punctuation at all here. What about substituting or adding `@partner.name`?

spec/controllers/hq/partners_controller_spec.rb
```diff
@@ -17,3 +17,28 @@ def foobar_action
...
 end
+
+RSpec.describe Hq::PartnersController, type: :controller do
+  before :each do
+    sign_in instance_double(AdminUser)
+    @partner = build_stubbed :partner
+    expect(Partner).to receive(:find).and_return(@partner)
+  end
+
+  it 'editing renders the edit view' do
```
NikitaAvvakumov: `specify` reads more naturally here.

spec/controllers/hq/partners_controller_spec.rb
```diff
@@ -17,3 +17,28 @@ def foobar_action
...
+    get :edit, id: @partner.id
+    expect(response).to render_template 'edit'
+  end
+
+  it 'unsuccessful update renders the edit view' do
+    expect(@partner).to receive(:save).and_return(false)
+    patch :update, id: @partner.id, partner: {name: ""}
```
NikitaAvvakumov: Rather out of curiosity - does `.and_return(false)` above make `partner: { name: "" }` unnecessary?

app/controllers/hq/partners_controller.rb
PurityControl: because its part of the build_partner method which will build a brand new partner for new actions where there are no params and assign params when the action is an update

app/views/hq/partners/edit.html.erb
PurityControl: which substituting or adding ?

app/assets/javascripts/partners.js.coffee
```diff
@@ -0,0 +1,5 @@
+jQuery ->
+  $("#partner_start_date").datepicker({
+    format: "yyyy-mm-dd"
+    autoclose: true
+  })
```
NikitaAvvakumov: Is there a way to generalize this so that we don't have to add JS files for other models' views with date fields?

spec/controllers/hq/partners_controller_spec.rb
PurityControl: I think this is left over from trying to make it fails before I stubbed - will fix it.

app/assets/javascripts/partners.js.coffee
PurityControl: are you thinking of the equivalent of a helper method to be called when needed?

spec/controllers/hq/partners_controller_spec.rb
```diff
@@ -17,3 +17,28 @@ def foobar_action
     end
   end
 end
+
+RSpec.describe Hq::PartnersController, type: :controller do
+  before :each do
+    sign_in instance_double(AdminUser)
+    @partner = build_stubbed :partner
+    expect(Partner).to receive(:find).and_return(@partner)
```
PurityControl: checked this as promised - if this line is removed then the tests fail

app/controllers/hq/partners_controller.rb
NikitaAvvakumov: Right, as per their example:
```ruby
# controller
def edit
  @aggregate = FacadeClass.new(get_partner)
end

private
def get_partner
  Partner.find(params[:id])
end

# facade_class.rb
class FacadeClass
  def initialize(partner)
    @partner = partner
  end

  def provinces
    Province.all
  end

  def statuses
    Status.all
  end
end

# _form.html.erb
@aggregate.partner.name # not a violation of Demeter!
@aggregate.provinces
@aggregate.statuses
```

I see value in a single "package" being passed from controller to the view

app/controllers/hq/partners_controller.rb
NikitaAvvakumov: Does that mean that `build_partner` is being called upon when it shouldn't be? Could you not have the `new` action simply assign `@partner = Partner.new` or call `build_new_partner`? Or `@aggregate = FacadeClass.new(Partner.new)` if that's the road we take.

app/views/hq/partners/edit.html.erb
NikitaAvvakumov: Sorry:
```erb
<h1> Edit Partner <%= @partner.name %>, OSRA number <%= @partner.osra_num %> </h1>
```

app/assets/javascripts/partners.js.coffee
NikitaAvvakumov: Would reusing the same CSS ids by removing the `partner_` part be bad?

spec/controllers/hq/partners_controller_spec.rb
razorcd: meaning that `@partner` defined here is accessed from the controller, as we discussed in the scrum?

app/controllers/hq/partners_controller.rb
PurityControl: I had actually considered it but, to me, assigns is about variables, where loads brings to attention stuff is being brought in from the database.
I'm comfortable with either.

app/assets/javascripts/partners.js.coffee
PurityControl: Yes very bad, wicked even. Id's should be unique and we are increasing the chance of a collision, especially across views using several partials.
However, I have renamed the file date-picker and it triggers on the css class "date-picker" so now people just have to add "date-picker" in their css class and they get the date-picker for free.

app/controllers/hq/partners_controller.rb
PurityControl: Skirting the aggregate issue for the moment because I would like to discuss that in more detail elsewhere.
No I don't think it is. We do the same things again and again in all of our controllers. We load, save and build our model ready for saving. 
Building is the same concept for the new, create and update methods. The only difference is their state. So for new we build a model without an associated database record and empty params, the same for create but we wan't to merge in form params, with an update we are looking at merging in form params to an existing record. This method covers all bases and is DRY.
 - build a new partner if one doesn't exist
 - merge in attributes if there are any
additionally Partner.new and Partner.build are just aliases https://github.com/rails/rails/blob/master/activerecord/lib/active_record/relation.rb#L124

app/controllers/hq/partners_controller.rb
theB264: If it's hitting the db, I think `load` is a good choice

NikitaAvvakumov:
@PurityControl Shall we leave the facade discussion & decision for another PR? Rebase & merge this one?

PurityControl:
Yeah can do, but I do want to come back to that because I have had similar thoughts and some ideas too. 
https://osraav.atlassian.net/browse/OSRA-322

PR #253, opened 2015-01-02
New default form builder incorporating errors

PurityControl:
create shared/errors partial for rendering form errors

`<%= render 'shared/errors', object: f.object %>`


PurityControl:
screen shot 2015-01-02 at 08 34 21

razorcd:
for the error messages  (Name can't be blank) would it look better with white background and red text?
So instead `class="bg-danger"` use  `class="text-danger"`
http://getbootstrap.com/css/#helper-classes-colors

PurityControl:
I think so.

PurityControl:
screen shot 2015-01-03 at 21 01 34

app/helpers/form_errors_builder.rb
```diff
@@ -0,0 +1,29 @@
...
+    @template.content_tag :div, class: "panel panel-danger" do
+      error_header << error_panel
+    end
+  end
+
+private
+  def list_of_errors
+    @template.content_tag(:ul, class: "text-danger") do
+      object.errors.full_messages.map { |message|
+        @template.content_tag(:li, message) }.join("\n").html_safe
```
NikitaAvvakumov: `do..end` instead of curly braces here?

app/helpers/form_errors_builder.rb
```diff
@@ -0,0 +1,29 @@
...
+  def error_messages
+    return unless object.respond_to?(:errors) && object.errors.any?
+
+    @template.content_tag :div, class: "panel panel-danger" do
+      error_header << error_panel
+    end
+  end
+
+private
+  def list_of_errors
```
NikitaAvvakumov: Prefer to define methods in order of calling - here `error_header`, `error_panel`, `list_of_errors`.

app/helpers/form_errors_builder.rb
```diff
@@ -0,0 +1,29 @@
...
+  def error_header
+    @template.content_tag :div, class: "panel-heading" do
+      @template.content_tag(:h3,
+        "Please fix the errors below and resubmit the form.",
+        class: "panel-title")
+    end
+  end
+
+  def error_panel
+      @template.content_tag(:div, list_of_errors, class: "panel-body")
```
NikitaAvvakumov: Please fix indent

app/helpers/form_errors_builder.rb
```diff
@@ -0,0 +1,29 @@
+class FormErrorsBuilder < ActionView::Helpers::FormBuilder
```
NikitaAvvakumov: What do we gain from this implementation at the cost of complexity over a simple shared errors partial such as this generic one https://github.com/NikitaAvvakumov/blog/blob/master/app/views/shared/_error_messages.html.erb?

spec/views/form_errors_builder_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+require 'cgi'
+
+RSpec.describe "hq/partners/_form.html.erb", type: :view do
+
+  let(:provinces) { Province.all }
```
NikitaAvvakumov: Any advantage to splitting up the setup into `let` & `before` instead of
```ruby
before :each do
  assign :provinces, Province.all
  assign :statuses, Status.all
  partner = build :partner, name: nil, region: 'Test'
  assign :partner, partner
  ..
end
```
(syntax differences not meant to be suggestions :smile: )

app/helpers/form_errors_builder.rb
PurityControl: Why would we go for the alternative just because of slightly reduced complexity (so soon after last nights facade pattern discussion :wink:)?
- much of the difference is html noise between a list and a bootstrap panel
- we have encapsulated all of the form handling for the object in a form_for model why would we put parts of it elsewhere?

spec/views/form_errors_builder_spec.rb
PurityControl: agreed - I'll change this

app/helpers/form_errors_builder.rb
PurityControl: fat fingers :disappointed:

app/helpers/form_errors_builder.rb
NikitaAvvakumov: I'm still not sold.
```erb
<!-- app/views/shared/_errors.html.erb -->
<% if object.respond_to(:errors) && object.errors.any? %>
  <div class="panel panel-danger">
    <div class="panel-heading">
      <h3 class="panel-title">
        Please fix the errors below and resubmit the form.
      </h3>
    </div>

    <div class="panel-body">
      <ul class="text-danger">
        <% object.errors.full_messages.each do |message| %>
          <li><%= message %></li>
        <% end %>
      </ul>
    </div>
  </div>
<% end %>

<!-- app/views/hq/<model>/_form.html.erb -->
<%= form_for [:hq, @model] do |f| %>
  <%= render 'shared/errors', object: f.object %>
  .
  .
<% end %>
```
 - the .erb file is almost half the size of the Ruby one
 - given that it's preparing (part of) a view to be rendered, the fact that the structure of that view is immediately visible seems of value
 - content tags are most useful when they interpolate CSS selectors and/or content but I see no utility in using them to generate static markup and largely static content. Overall, this applies to Ruby vs HTML - doesn't the latter feel like the right tool for the job?
 - there seems to be no difference in terms of encapsulation - the form either calls a helper or renders a partial

theB264: As much as I like encapsulation and OO design, Nik's suggestion seems much more readable to someone unfamiliar with the codebase

PurityControl: I take the criticism that the method has been poorly implemented in terms of putting html in there when that is not its responsibility. I think passing this on to a view when we have encapsulated every thing else about the form into a form helper is just wrong, but I don't see us reaching agreement on that side of things.

theB264:
:+1:

NikitaAvvakumov:
I would very much like to arrive at a consensus here instead of forcing something through.

Encapsulation and OO design are valuable concepts indeed, but perhaps their pursuit ought to be tempered by the practicalities. The Rails form helper is valuable because it allows one to write so much less boilerplate HTML and not worry about HTML syntax. It is intelligent enough to understand the minimal info provided to it and build it up into a fully bloated HTML equivalent. It buys us a lot (though Formtastic or Simple Form buy even more). 

By contrast, the FormErrorsBuilder is essentially a 1:1 representation of the HTML structure in Ruby - there is no logic, minimal querying of the underlying model(s), no interpolation of CSS selectors, no addition of hidden tag attributes. The ultimate job of both builders is to create an HTML view, but while the FormBuilder does a lot behind the scenes, the FormErrorsBuilder simply delivers a verbatim translation of Ruby to HTML.

In its simplest form, the FormErrorsBuilder can be written as
```ruby
class FormErrorsBuilder < ActionView::Helpers::FormBuilder
  def error_messages
    return unless object.respond_to?(:errors) && object.errors.any?

    @template.content_tag :div, class: 'panel panel-danger' do
      @template.content_tag :div, class 'panel-heading' do
        @template.content_tag :h3,
          'Please fix the errors below and resubmit the form.',
          class: 'panel-title'
      end
    end

    @template.content_tag :div, class: 'panel-body' do
      @template.content_tag :ul, class: 'bg-danger' do
        object.errors.full_messages.each do |msg|
          @template.content_tag :li, msg
        end
      end
    end
  end
end
```

This is, of course, the exact structure of the erb partial. At least at present, there is no obvious reason to over-complicate it.

The above can be rewritten as
```ruby
class FormErrorsBuilder < ActionView::Helpers::FormBuilder
  def error_messages
    return unless object.respond_to?(:errors) && object.errors.any?

    @template.content_tag :div, class: 'panel panel-danger' do
      error_header
      error_body
    end
  end

  private

  def error_header
    @template.content_tag :div, class 'panel-heading' do
      @template.content_tag :h3,
        'Please fix the errors below and resubmit the form.',
        class: 'panel-title'
    end
  end

  def error_body
    @template.content_tag :div, class: 'panel-body' do
      @template.content_tag :ul, class: 'bg-danger' do
        error_list
      end
    end
  end

  def error_list
    object.errors.full_messages.each do |msg|
      @template.content_tag :li, msg
    end
  end
end
```
This is essentially @PurityControl's original, with minor changes for readability. It's better Ruby code than the first snippet, but this improvement comes at the expense of readability and clarity of structure of the final HTML. It would make sense to adopt it if the modularity was useful to us in easily changing and extending the 3 private methods. However, until that becomes of value, the first snippet is simply clearer. And if we go with it, then we might as well take the direct route of writing HTML in HTML.

Or, as @theB264 said with 1% of the words, "As much as I like encapsulation and OO design, <the erb partial> seems much more readable".

PurityControl:
In isolation of the error messages much of what you say I agree with, including maintenance of the html. The question then becomes once we have implemented a stable design how often will we need to change the html.
But if the helper was extended I envision being able to write the following for all our forms. 
```
<%= form_for [:hq, @partner] do |f| %>
  <%= f.error_messages %>
  <%= f.osra_text_field :name %>
  <%= f.osra_text_field :region %>
  <%= f.osra_text_field :contact_details %>
  <%= f.osra_collection_select :province_id, @provinces, :id, :name , {} %>
  <%= f.osra_collection_select :status_id, @statuses, :id, :name ,  {} %>
  <%= f.osra_text_field :start_date %>
  <%= f.submit %>
<% end %>
```

But really , I'm okay with the majority decision here.

NikitaAvvakumov:
It seems, then, that we should be interested in Simple Form or Formtastic.

PR #255, opened 2015-01-03
URGENT fix for osra orphan import requested by client

PurityControl:
- removed mandatory status on attributes
  - minor_siblings_count
  - original neighbourhood
  - current neighbourhood
  - guardian id num
- validations that just need to be removed have been only commented out given
  that the request suggest this is only temporary
- no major attempt to refactor minor sibling count cross validation again given
  the temporary nature of the issue
- additional tests for the cross validation of sibling count
- changed config/settings.yml to remove mandatory status from fields

### DIFF
--- a/app/models/orphan.rb
+++ b/app/models/orphan.rb
@@ -38,7 +38,7 @@ class Orphan < ActiveRecord::Base
   validates :gender, presence: true, inclusion: {in: Settings.lookup.gender }
   validates :contact_number, presence: true
   validates :sponsored_by_another_org, inclusion: {in: [true, false] }, exclusion: { in: [nil]}
-  validates :minor_siblings_count, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+  validates :minor_siblings_count, numericality: { only_integer: true, greater_than_or_equal_to: 0 },  allow_nil: true
   validates :sponsored_minor_siblings_count, numericality: { only_integer: true, greater_than_or_equal_to: 0 }, allow_nil: true
   validate :sponsored_siblings_does_not_exceed_siblings_count
   validates :original_address, presence: true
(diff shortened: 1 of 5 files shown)


app/models/orphan.rb
```diff
@@ -127,6 +127,12 @@ def current_sponsor
...
   def sponsored_siblings_does_not_exceed_siblings_count
+    if minor_siblings_count.nil?
+      unless sponsored_minor_siblings_count == 0 or sponsored_minor_siblings_count.nil?
+        errors.add(:sponsored_minor_siblings_count, "cannot exceed minor siblings count")
+      end
+      return
+    end
     if sponsored_minor_siblings_count && (sponsored_minor_siblings_count > minor_siblings_count)
       errors.add(:sponsored_minor_siblings_count, "cannot exceed minor siblings count")
     end
```
NikitaAvvakumov:
```ruby
def sponsored_siblings_does_not_exceed_siblings_count
  return if sponsored_minor_siblings_count.nil? || sponsored_minor_siblings_count == 0
  if minor_siblings_count && sponsored_minor_siblings_count > minor_siblings_count
      errors.add(:sponsored_minor_siblings_count, 'cannot exceed minor siblings count')
  end
end
```

PurityControl: what about the minor_sibling_count.nil? guard - thats the crucial thing. 
Given neither are compulsory now the minor siblings_count could be nil while the sponsored_minor_siblings_count could have a value

PurityControl: additionally the first guard returns without an error

spec/models/orphan_spec.rb
```diff
@@ -142,6 +142,26 @@
       orphan.sponsored_minor_siblings_count = nil
       expect(orphan).to be_valid
     end
+
+    describe 'if sibling count nil then sponsored sibling count is nil or 0' do
```
NikitaAvvakumov: Prefer 
```ruby
context 'when minor_sibling_count is nil' do
it 'is valid when sponsored_minor_siblings_count is nil'
it 'is valid when sponsored_minor_siblings_count is 0'
it 'is not valid when sponsored_minor_siblings_count is greater than 0'
```

PurityControl: agreed - on it

app/models/orphan.rb
NikitaAvvakumov: Right!
```ruby
if minor_siblings_count.nil? || sponsored_minor_siblings_count > minor_siblings_count
  errors.add
```

PurityControl: if minor_siblings_count is nil then it fails based on the above but it should not fail when nil and either
- sponsored_minor_siblings_count is nil
- sponsored_minor_siblings_count is 0

PR #257, opened 2015-01-03
OSRA-321 Clean up importer error messages

NikitaAvvakumov:
screen shot 2015-01-03 at 7 45 03 pm

https://osraav.atlassian.net/browse/OSRA-321

 - clean up the call to `add_import_errors` within `check_orphan_validity`
   in orphan_importer.rb
 - improve error message for non-unique orphan records

### DIFF
--- a/app/models/orphan.rb
+++ b/app/models/orphan.rb
@@ -15,7 +15,7 @@ class Orphan < ActiveRecord::Base
 
   validates :name, presence: true,
             uniqueness: { scope: [:family_name, :mother_name, :father_given_name],
-                          message: 'An orphan with this name, father, mother & family name is already in the database.' }
+                          message: 'taken: an orphan with this name, father, mother & family name is already in the database.' }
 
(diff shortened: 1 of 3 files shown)


app/models/orphan.rb
```diff
@@ -15,7 +15,7 @@ class Orphan < ActiveRecord::Base
 
   validates :name, presence: true,
             uniqueness: { scope: [:family_name, :mother_name, :father_given_name],
-                          message: 'An orphan with this name, father, mother & family name is already in the database.' }
+                          message: 'taken: an orphan with this name, father, mother & family name is already in the database.' }
```
NikitaAvvakumov: The error message is automatically prepended with the attribute name - "Name" in this case. It is possible to override this behaviour, but only in a way that would affect all error messages. Changing this message from "Name An orphan with this name..." to "Name taken: an orphan with this name..." seems like an acceptable compromise.

PR #260, opened 2015-01-04
OSRA318 date format helper

theB264:
https://osraav.atlassian.net/browse/OSRA-318


spec/helpers/hq/time_helper_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'rails_helper'
+
+RSpec.describe Hq::TimeHelper, type: :helper do
+  it 'formats a date' do
+    expect(Hq::TimeHelper.date_to_string(Date.current)).to be_a_kind_of String
+  end
+
+  it "doesn't raise exception on nil" do
```
PurityControl: @NikitaAvvakumov is this not a primary case for a null object?

NikitaAvvakumov: So, something like
```ruby
# app/helpers/application_helper.rb
module ApplicationHelper
  def date_to_string(date = NullDate.new)
    date.strftime('%d %B %Y')
  end
end

# ??/null_date.rb
class NullDate
  def strftime(*args)
    'none'
  end
end
```

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,5 @@
 module ApplicationHelper
+  def date_to_string(date)
```
NikitaAvvakumov: Looking ahead, since we will need another helper that converts dates to mm/yyyy, can this method name be more explicit?

theB264: Nice catch

spec/helpers/application_helper_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'rails_helper'
+
+RSpec.describe ApplicationHelper, type: :helper do
+  context 'date_to_string' do
+    it 'formats a date' do
+      expect(date_to_string(Date.current)).to be_a_kind_of String
```
NikitaAvvakumov: Is there value in testing that the method not merely returns a string but a correctly formatted date? i.e.
```ruby
expect(date_to_string(Date.current)).to eq Date.current.strftime('%d %B %Y')
```
It's very redundant, but the current test doesn't seem to give much.

theB264: I pondered this but in the spirit of having one place where the format was specified, I opted not to do this.  Do you think it's better to have the format string appear twice?  I also considered stubbing strftime, would that make more sense here?

NikitaAvvakumov: Would you consider hard-coding a date here?

theB264: As long as the class is Date not String it's on the table

NikitaAvvakumov:
```ruby
expect(date_to_string(Date.parse('05 Jan 2015'))).to eq '05 January 2015'
```
?

spec/helpers/hq/time_helper_spec.rb
PurityControl: yes the only remaining question is what happens when someone calls it with something that is not a date?
Do we want it to blow up or do we want to return a Null Date?
If we want to return a null date in such circumstances do we want the message to be an argument to null date instantiation with a default ?

theB264: The newest commit returns nil if the argument is not an instance of `Date`

spec/helpers/application_helper_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
+
+RSpec.describe ApplicationHelper, type: :helper do
+  context ':format_full_date' do
+    it 'formats a date' do
+      allow_any_instance_of(Date).to receive(:strftime).and_return '45 Septemuary 1978'
+      expect(format_full_date(Date.current)).to eq '45 Septemuary 1978'
```
NikitaAvvakumov: Still unsure about these tests. They only confirm that the methods call `strftime` but not that they actually work as expected. The fact that they pass the right arguments for correct formatting seems at least as important.

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,15 @@
...
+  end
+
+  def format_month_date(date)
+    format date, '%m %Y'
+  end
+
+private
+
+  def format(date, format)
+    date.strftime(format) if date.is_a?(Date)
```
NikitaAvvakumov: You are not sold on the Null Object idea, then?

spec/helpers/application_helper_spec.rb
theB264: This was by design.  I was testing that a date was formatted, not formatted a certain way, in order to DRY out the format process and create one authoritative source for the way dates are diplayed.  Do you think it's better to create a constant for the format string, test that it is set, and use the constant in the tests?  I pondered this but really liked it this way.  That seems like overkill here, but if the team thinks it's better I'll send it up

app/admin/sponsor.rb
```diff
@@ -75,7 +75,7 @@
         column :orphan_date_of_birth
         column :orphan_gender
         column 'Sponsorship began' do |_sponsorship|
-          _sponsorship.start_date.strftime("%m/%Y")
+          ApplicationHelper.instance_method(:format_month_year_date).bind(binding).call(_sponsorship.start_date)
```
NikitaAvvakumov: That looks painful. What do you think about letting AA live out its remaining days as is? Or creating a separate helper for it that will be deleted along with the rest of AA code when time comes?

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,10 @@
 module ApplicationHelper
+  FULL_DATE_FORMAT_STRING= '%d %B %Y'
+  MONTH_YEAR_DATE_FORMAT_STRING= '%m/%Y'
+
+  ['full', 'month_year'].each do |type_of|
```
NikitaAvvakumov: I think this gives up too much readability for the sake of DRYness.

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,10 @@
...
+    define_method "format_#{type_of}_date" do |date|
+      date.try :strftime, %Q~#{ApplicationHelper.const_get("#{type_of.upcase}_DATE_FORMAT_STRING".to_sym)}~
```
NikitaAvvakumov: `.try` is considered a code smell as it effectively says "I have no idea what's going on here but I'll give it a shot". Here, specifically, I think it also can be harmful by letting callers of this method get away with too much: The method is clearly about time conversion. There may be times when it's called with `nil` if a particular date attribute is missing. Using a true Null Object helps to avoid type checking in the views in such cases, which is great - we don't have to write
```erb
<% if @sponsorship.end_date %>
  <%= format_month_year_date @sponsorship.end_date %>
<% end %>
```
However, allowing the method to be called with _any_ object type without blowing up can lead to sloppiness and difficult-to-track bugs. With that in mind, I still prefer the Null Object implementation we'd come up with in the comments earlier.

spec/helpers/application_helper_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  context ':format_full_date' do
+    it 'formats a date' do
+      expect_any_instance_of(Date).to receive(:strftime).
+                                  with(ApplicationHelper::FULL_DATE_FORMAT_STRING).
+                                  and_return '45 März 1776'
+      expect(format_full_date(Date.current)).to eq '45 März 1776'
+    end
+
+    it "doesn't raise exception on nil" do
```
NikitaAvvakumov: Echoing my earlier comment: the method doesn't raise exceptions no matter what object type it receives, which makes this test somewhat disingenuous.

spec/helpers/application_helper_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      end
+
+      it 'is a String' do
+        expect(ApplicationHelper::MONTH_YEAR_DATE_FORMAT_STRING).to be_a_kind_of String
+      end
+    end
+  end
+
+  context ':format_full_date' do
+    it 'formats a date' do
```
NikitaAvvakumov: I'd like to get the team's opinion here. My feeling is that this spec mocks too much. At least one scenario where this testing is insufficient involves the constants: if they are not set properly - e.g. `'%b%Y  -%D'` instead of `'%d %B %Y'` - the test will still pass but the actual output will be a mess. Never mind a situation where Ruby's `.strftime` changes.

My preference here would for a more functional test - verifying that given an actual date, the method returns its properly formatted string representation.

theB264: Constants are now tested that they are set properly, at the expense of having a single authoritative place where date formats are set.  If `.strftime` changes a deprecation warning should be issued.

spec/helpers/application_helper_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'rails_helper'
+
+RSpec.describe ApplicationHelper, type: :helper do
+  describe 'date formatting constants' do
+    context 'FULL_DATE_FORMAT_STRING' do
+      it 'exists' do
+        expect(ApplicationHelper::FULL_DATE_FORMAT_STRING).to be_present
+      end
+
+      it 'is set' do
```
PurityControl: if its set, surely it exists. Is the previous test necessary?

spec/helpers/application_helper_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+        expect(ApplicationHelper::FULL_DATE_FORMAT_STRING).to eq '%d %B %Y'
+      end
+    end
+
+    context 'MONTH_YEAR_DATE_FORMAT_STRING' do
+      it 'exists' do
+        expect(ApplicationHelper::MONTH_YEAR_DATE_FORMAT_STRING).to be_present
+      end
+
+      it 'is set' do
```
PurityControl: as above

spec/helpers/application_helper_spec.rb
PurityControl: I don't share your worry about `.strftime` suddenly changing from under us. I do agree with you on what we are actually testing though. This reminds me of tests that check a certain function have been called, in a slightly different format and seems tightly coupled to its implementation.
I would expect to see a test along the format of.
- Create a Date object  with a specific day, month, year
- check its format matches what we expect for that date

NikitaAvvakumov:
I'd like to see the tests of constants removed and the commit history squashed, but otherwise this PR is ready to go.

PR #262, opened 2015-01-04
[WIP] OSRA-299 Atomic sponsorship inactivation

NikitaAvvakumov:
https://osraav.atlassian.net/browse/OSRA-299

All changes associated with sponsorship inactivation wrapped in an AR
transaction to ensure that either all go through or none:

 - `sponsorship.update_attributes!`, `set_orphan_status_to_previously_sponsored`
   and `update_sponsor` are called from within `ActiveRecord::Base.transaction`
   block in `Sponsorship#inactivate`
 - any exceptions raised by the above methods are rescued by adding an error
   message to the sponsorship object and returning `false` to the controller
 - `set_orphan_status_to_sponsored` & `update_sponsor` moved to `after_create`

### DIFF
--- a/spec/models/sponsorship_spec.rb
+++ b/spec/models/sponsorship_spec.rb
@@ -34,7 +34,7 @@ describe Sponsorship, type: :model do
     describe 'start_date' do
       before(:all) { travel_to Date.parse "15-12-2011" }
       after(:all) { travel_back }
-      
+
       let (:today) { Date.current }
       let (:yesterday) { today.yesterday }
       let (:first_of_next_month) { today.beginning_of_month.next_month }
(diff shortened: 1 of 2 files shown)


NikitaAvvakumov:
Since `save` & `destroy` are wrapped in AR transactions by default, an alternative way to handle this would be to call sponsor & orphan update methods from the `after_save` hook. `update_sponsor` is already intelligent enough for this. However, the orphan update is currently split into two methods - `set_orphan_status_to_sponsored` & `set_orphan_status_to_previously_sponsored`. If sponsorship were to simply call `update_orphan` instead, and the Orphan model then figured which status it needs to set, then we could do the following:

```ruby
# app/models/sponsorship.rb

# after_create :update_sponsor, :set_orphan_status_to_sponsored
after_save :update_sponsor, :update_orphan

def inactivate(end_date)
  update_attributes!(active: false, end_date: end_date)
end
```
This implementation looks cleaner to me _and_ transfers the logic and responsibility from Sponsorship to Orphan where I think it belongs.

Thoughts?

spec/models/sponsorship_spec.rb
```diff
@@ -128,6 +128,47 @@
...
       end
+
+      describe 'atomicity' do
+        let(:sponsor) { create :sponsor }
+        let(:orphan) { create :orphan }
+        let!(:sponsorship) do
+          create :sponsorship, sponsor: sponsor, orphan: orphan
+        end
+
+        def relevant_attributes
```
PurityControl: why a method rather than a let?

NikitaAvvakumov: Because `let` memoizes while the method resamples

PurityControl:
This brings up the issue of refactoring Orphan. 

We have talked about tidying up father and how to tease out databases columns that are part of the table to do this. 

However there is another substantial part of the class that deals with sponsorship issues that orphan shouldn't have to know or worry about. 

I think we are missing an additional class here that has been left out of the domain say OrphanSponsorship for lack of imagining any better description at the moment. This additional class would be ideally suited to taking some of the sponsorship weight off Orphan's shoulders.

But I do not think sponsorship logic should be passed on to either the orphan or sponsor models.

NikitaAvvakumov:
Extract a service object that handles sponsorship-related state changes for orphan & sponsor, and implement the observer pattern whereby sponsorship's state change notifies it to execute?
(It's very cool that I actually understand what I just said!)

I would very much like to combine or precede this with converting orphan's statuses to enums for state-machine-like behaviour.

NikitaAvvakumov:
Will wait until model refactoring begins

PR #264, opened 2015-01-06
Osra319 partners show action items

theB264:
https://osraav.atlassian.net/browse/OSRA-319

### DIFF
--- a/app/views/hq/partners/show.html.erb
+++ b/app/views/hq/partners/show.html.erb
@@ -4,6 +4,10 @@
                 class: 'btn btn-default', role: 'button') %>
     <%= link_to('Upload Orphan List', upload_hq_partner_pending_orphan_lists_path(@partner.id),
                 class: 'btn btn-primary', role: 'button') %>
+    <%- unless @partner.orphan_lists.empty? -%>
+      <%= link_to('Orphan lists', hq_partner_orphan_lists_path(@partner),
+                class: 'btn btn-default', role: 'button') %>
+    <%- end -%>
   </div>
 </div>
 <dl class="dl-horizontal">
@@ -33,14 +37,6 @@
 <dl class="dl-horizontal">
   <dt>Contact Details</dt>
   <dd><%= @partner.contact_details %></dd>
-</dl>
-<dl class="dl-horizontal">
-  <dt>Orphan Lists</dt>
-  <dd><%- if @partner.orphan_lists.empty? %>
-    <i>none</i>
-  <%- else %>
-    <%= link_to('All orphan lists', hq_partner_orphan_lists_path(@partner)) %>
-  <%- end %></dd>
 </dl>
 <dl class="dl-horizontal">
   <dt>Created at</dt>
--- a/app/views/hq/partners/show.html.erb
+++ b/app/views/hq/partners/show.html.erb
@@ -1,3 +1,11 @@
+<div class="action_items">
+  <div class="instance">
+    <%= link_to('Edit Partner', edit_hq_partner_path(@partner.id),
+                class: 'btn btn-default', role: 'button') %>
+    <%= link_to('Upload Orphan List', upload_hq_partner_pending_orphan_lists_path(@partner.id),
+                class: 'btn btn-primary', role: 'button') %>
+  </div>
+</div>
 <dl class="dl-horizontal">
   <dt>Partner Name</dt>
   <dd><%= @partner.name %></dd>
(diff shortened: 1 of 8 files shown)


theB264:
screenshot from 2015-01-06 02 56 00

spec/routing/admin/pending_orphan_lists_routing_spec.rb
```diff
@@ -10,4 +10,6 @@
   specify { expect(put: '/admin/partners/1/pending_orphan_lists/1').not_to be_routable }
   specify { expect(patch: '/admin/partners/1/pending_orphan_lists/1').not_to be_routable }
   specify { expect(delete: '/admin/partners/1/pending_orphan_lists/1').to be_routable }
+
+  #missing: (post: '/admin/partners/1/pending_orphan_lists') Admin::PendingOrphanListsController#create
```
theB264: OSRA346
https://osraav.atlassian.net/browse/OSRA-346

config/routes.rb
```diff
@@ -3,6 +3,9 @@
   namespace :hq do
     resources :partners, except: [:destroy] do
       resources :orphan_lists, only: [:index]
+      resources :pending_orphan_lists, only: [:create, :new, :destroy] do
+        ['upload', 'validate', 'import'].each { |action| get action, on: :collection }
```
PurityControl: the get is hidden in the logic, i know its more verbose but I find it easier to understand
```
collection do
  get :upload
  get :validate
  get :import
end
```
additionally once uploaded are the `validate` and `import` actions member actions ?

app/views/hq/partners/show.html.erb
```diff
@@ -1,3 +1,11 @@
+<div class="action_items">
+  <div class="instance">
+    <%= link_to('Edit Partner', edit_hq_partner_path(@partner.id),
+                class: 'btn btn-default', role: 'button') %>
+    <%= link_to('Upload Orphan List', upload_hq_partner_pending_orphan_lists_path(@partner.id),
```
NikitaAvvakumov: What do you think about simplifying this route helper?

config/routes.rb
NikitaAvvakumov: So, this generates 6 routes for `pending_orphan_lists` in total? Is `:create` used?

NikitaAvvakumov:
This PR seems to go well beyond its intended scope, with most of its code dedicated to pending orphan lists, not partner. Can we just use a stub route ('#') for the Upload button for now and come back to pending orphan lists in due time?

app/views/hq/partners/show.html.erb
theB264: The router only allows you to override the current prefix, so
`get 'upload', on: :collection, as: :foobar` is `foobar_hq_partner_pending_orphan_lists_path` instead of `upload_hq_partner_pending_orphan_lists_path`.  The alternative is to hardcode the URL outside of the blocks & namespaces like `:new_sponsorship` is, which I think would be more messy and harder to understand

theB264:
The way I understood the story was to create the "action items" and the "action" part of that is mapping the link to a controller method and using a path helper for linking.

config/routes.rb
theB264: Nice catch.  `:create` and `:new` appear to be unused in the old code

config/routes.rb
```diff
@@ -3,6 +3,12 @@
   namespace :hq do
     resources :partners, except: [:destroy] do
       resources :orphan_lists, only: [:index]
+      resources :pending_orphan_lists do
+        delete 'destroy', on: :member
+        get 'upload', on: :collection
+        get 'validate', on: :collection
```
PurityControl: are `validate` and `upload` `collection` or `member` resources ?

theB264: They are `collection` resources

spec/views/hq/partners/show_spec.rb
```diff
@@ -5,11 +5,23 @@
...
+      expect(rendered).to have_link('Edit Partner', edit_hq_partner_path(partner.id))
     end
 
+    specify 'Upload Orphan List' do
+      render
+      expect(rendered).to have_link('Upload Orphan List', upload_hq_partner_pending_orphan_lists_path(partner.id))
+    end
+  end
+
+  describe 'orphan lists link' do
```
NikitaAvvakumov: This is an instance action-item too - move the tests into the block?

theB264: Nice catch

PR #265, opened 2015-01-06
osra323 partners#index action item

theB264:
https://osraav.atlassian.net/browse/OSRA-323
screenshot from 2015-01-06 03 24 51

### DIFF
--- a/app/views/hq/partners/index.html.erb
+++ b/app/views/hq/partners/index.html.erb
@@ -1,47 +1,52 @@
-
-    <%- if @partners.empty? %>
-      No Partners found
-    <%- else %>
-      <table>
-        <thead>
-          <tr>
-            <th>
-              Osra Num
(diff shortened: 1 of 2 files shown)


app/views/hq/partners/index.html.erb
```diff
@@ -1,47 +1,52 @@
...
-        </tbody>
-      </table>
-    <%- end %>
+<div class="action_items">
+  <div class="class">
+    <%= link_to('New Partner', new_hq_partner_path,
+                class: 'btn btn-default', role: 'button') %>
+  </div>
+</div>
+<%- if @partners.empty? %>
```
PurityControl: is this guard worth the trouble?

theB264: This line wasn't changed, I just fixed the indent for the entire block

PR #270, opened 2015-01-09
OSRA-310 Upgrade to Ruby 2.2.0

NikitaAvvakumov:
OSRA-310 Upgrade to Ruby 2.2.0

https://osraav.atlassian.net/browse/OSRA-310

 - update Gemfile to specify `ruby '2.2.0'`
 - update cucumber-rails to include cucumber v1.3.8 for 2.2.0 compatibility
   (this updates several dependencies, including Rails from 4.1.8 to 4.1.9)
 - update ruby-ole gem for ruby 2.2.0 compatibility

## Bug revealed:
On L15 of app/models/sponsorship.rb
`validates :end_date, presence: {scope: true, message: 'is invalid' }, if: '!active', if: 'active_changed?'`
the duplicate `if` key is ignored, i.e. only the first conditional is used for validating `end_date`.
Fix task is OSRA-347

PR #271, opened 2015-01-09
Osra 347 fix sponsorship validation conditional

NikitaAvvakumov:
OSRA-347 Fix sponsorship.rb end_date validation

https://osraav.atlassian.net/browse/OSRA-347

_This PR is based on #270, but can be rebased on develop if needed_

 - replace double conditional with single `if: '!active'` - end date must
   always be present when sponsorship is inactive, not only at the moment
   the inactivation takes place
 - add `nil` as an unacceptable value for inactive sponsorship's `end_date`
   in sponsorship_spec.rb

### DIFF
--- a/app/models/sponsorship.rb
+++ b/app/models/sponsorship.rb
@@ -12,7 +12,7 @@ class Sponsorship < ActiveRecord::Base
   validates :start_date, presence: { scope: true, message: "is invalid"}
   validate  :start_date_no_later_than_1st_of_next_month, if: :start_date
   
-  validates :end_date, presence: {scope: true, message: 'is invalid' }, if: '!active', if: 'active_changed?'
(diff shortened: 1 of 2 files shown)


NikitaAvvakumov:
Build will keep failing until Semaphore is configured to use Ruby 2.2.0, but everything is passing on my repo/Semaphore instance.

PurityControl:
What value is the active attribute when a sponsorship is set in the future but hasn't started yet?

NikitaAvvakumov:
We don't have that kind of logic yet. As soon as it's created, the sponsorship is `active: true`, regardless of `start_date`. Same with inactivation.

PurityControl:
do you want to rebase this ?

NikitaAvvakumov:
Rebased & green

PR #272, opened 2015-01-10
Update Rails and other gems

NikitaAvvakumov:
- update Rails to 4.2.0
   - add 'web-console' gem for debugging console in browser
   - 4.2.0 allows implementation of foreign key constraints (OSRA-146)
 - allow ActiveAdmin to work with Rails 4.2.0
   - source 'inherited_resources' from 'rails-4-2' branch for 4.2.0 compatibility
 - update all gems
 - handle 4.2.0 deprecation warnings
   - opt in to propagate errors raised in `after_rollback` & `after_commit`
   - rename `serve_static_assets` to `serve_static_files`
 - fix one failing Cucumber scenario & 2 failing specs


features/admin/orphan_lists.feature
```diff
@@ -158,7 +158,7 @@ Feature:
     Then I click the "Upload" button
     Then I should see "محمد"
     And I should see "John Doe"
-    And I should see "January 02, 2012"
+    And I should see "2012-01-02"
```
PurityControl: Why does the rails upgrade make this fail?

theB264: Is this related to https://github.com/AgileVentures/osra/pull/260?  I remember the client said they want day, month, year everywhere so I made that global, even across the AA side.  But, I didn't see any tests fail

Gemfile
```diff
@@ -15,6 +15,8 @@ gem 'bootstrap-datepicker-rails'
 gem 'autoprefixer-rails', '~> 4.0.2'
 gem 'coffee-rails', '~> 4.1.0'
 gem 'uglifier', '>= 2.6.0' # compressor for JavaScript assets
+# activeadmin dependency
+gem 'inherited_resources', github: 'josevalim/inherited_resources', branch: 'rails-4-2'
```
PurityControl: do the benefits of upgrading now outweigh waiting for a production release?

theB264:
ActiveAdmin does not support Rails 4.2 yet.

features/admin/orphan_lists.feature
NikitaAvvakumov: This is in app/views/admin/pending_orphan_lists/valid_list.html.arb. It didn't receive the date formatter and uses simply `column :date_of_birth`. Best guess is that default date formatting changed somewhere after gems upgrade.

Gemfile
NikitaAvvakumov: Probably not. My initial intent was to upgrade only the outdated gems highlighted by Gemnasium. Rails just happened as I got carried away.

Also @theB264: AA can work with 4.2, their new lock is < 5.0. However, AA's dependency inherited_resources is not ready yet for 4.2 on master, which is why it's sourced from a feature branch. For what it's worth, there is a disclaimer on the [inherited_resources' repo](https://github.com/josevalim/inherited_resources), presumably from its creator Jose Valim, which reads
> Since Rails 3 came out, I have no longer used Inherited Resources and it is no longer actively maintained.

The repo is somewhat active, though most of that activity appears to be maintenance. So, master branch or feature, we will be/are using a pretty stale dependency.

Let me know if rollback to 4.1.9 is preferred (we do need to upgrade from 4.1.8 to 4.1.9 to enable other gems' updates).

theB264: `rollback to 4.1.9 is preferred== true`

NikitaAvvakumov:
Will wait until inherited_resource master branch works with rails 4.2.0

PR #273, opened 2015-01-10
SimpleCov configuration

NikitaAvvakumov:
- disable combined RSpec & Cucumber coverage reporting - will report RSpec only
 - exclude app/admin folder from coverage reporting
 - configure SimpleCov & Coveralls to both send coverage data to Coveralls
   and to create a report locally

At the end or running the RSpec suite, receive message `Coverage report generated for RSpec to ./coverage. 609 / 618 LOC (98.54%) covered.`

Can open local file ./coverage/index.html to see
screen shot 2015-01-08 at 4 59 09 pm
from which line-by-line coverage can be viewed by clicking on individual file names.
_(98.54% is indeed our current RSpec test coverage of everything outside the app/admin folder)_


theB264:
:shipit:

PR #274, opened 2015-01-10
OSRA-331 & 337 Rails routes/helpers and #index-related functionality for Users completed

robbiejackson:
Included routes for Users, controller and view functionality for #index, and related rspec tests.

### DIFF
--- a/spec/views/hq/users/index_spec.rb
+++ b/spec/views/hq/users/index_spec.rb
@@ -1,0 +1,52 @@
+require 'rails_helper'
+
+describe "hq/users/index.html.erb", type: :view do
+
+  it 'should have a New User link' do
+    assign(:users, [])
+    render
+    expect(rendered).to have_link('New User', new_hq_user_path)
+  end
+
+  describe 'should display users correctly' do
+
+    context 'no users' do
+      it 'displays just the header row' do
+        @users = []
+        render
+        expect(rendered).to have_selector('table tr', count: 1)
+      end
+    end
+
+    context '1 user' do
+      it 'displays a single user' do
+        @users = [ double('User', user_name: 'Frederick Bloggs', 
+          email: 'fred.bloggs@example.com', active_sponsors: (1..5).to_a)]
+        render
+        expect(rendered).to have_selector('table tr', count: 2)
+        expect(rendered).to have_link('Frederick Bloggs', href: hq_user_path(@users[0]))
+        expect(rendered).to match 'fred.bloggs@example.com'
+        expect(rendered).to match '5'
+      end
+    end
+
+    context '>1 user' do
+      it 'displays multiple users' do
+        @users = [ double('User', user_name: 'Frederick Bloggs', 
+          email: 'fred.bloggs@example.com', active_sponsors: (1..5).to_a), 
+                   double('User', user_name: 'Joseph Soap', 
+          email: 'joe.soap@example.com', active_sponsors: (1..7).to_a) ]
+        render
+        expect(rendered).to have_selector('table tr', count: 3)
+        expect(rendered).to have_link('Frederick Bloggs', href: hq_user_path(@users[0]))
+        expect(rendered).to match 'fred.bloggs@example.com'
+        expect(rendered).to match '5'
+        expect(rendered).to have_link('Joseph Soap', href: hq_user_path(@users[1]))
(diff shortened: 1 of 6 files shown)


spec/views/hq/users/index_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    expect(rendered).to have_link('New User', new_hq_user_path)
+  end
+
+  describe 'should display assigned active sponsors correctly' do
+
+    context 'no active sponsors' do
+      it 'displays 0 active sponsors' do
+        @users = [ instance_double(User, active_sponsors: [], user_name: 'usr', email: 'email@e.e') ]
+        render
+        expect(rendered).to have_selector('td', text: '0')
```
NikitaAvvakumov: @robbiejackson Having given this more thought after our conversation, on the one hand I think that this is still unnecessarily overstepping the boundaries of a focused unit test. The point is to test the view's successful rendering of data it is given, and not of any underlying logic since that is tested elsewhere. With that in mind, 'Assigned Active Sponsors' is no different than other attributes of the instance variable passed to the view and does not need to be handled specially. To test the rendered page, then, all we need is
```ruby
it 'should display users' do
  user = instance_double User, :user_name => 'H. Vetinari', 
                               :email => 'vetinari@a-m.gov', 
                               :active_sponsors => (1..5).to_a
  assigns :users, [user]

  render

  expect(rendered).to match /H\. Vetinari/
  expect(rendered).to match /vetinari@a-m\.gov/
  expect(rendered).to match '5'
end
```

On the other hand, if my opinion is not shared by the rest of the team and the test is kept as is, I would recommend making it precise and ensuring that the sponsors' count is displayed in the correct table column, not simply in a `'td'`.

spec/views/hq/users/index_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    context '>0 active sponsors' do
+      it 'displays >0 active sponsors' do
+        @users = [ instance_double(User, active_sponsors: ['s1','s2'], user_name: 'usr', email: 'email@e.e') ]
+        render
+        expect(rendered).to have_selector('td', text: '2')
+        expect(rendered).not_to have_selector('td', text: '0')
+      end
+    end
+ 
+=begin
```
NikitaAvvakumov: While I definitely agree that there is value in discussing integration testing, let's remove this code from here before merging.

PurityControl: agreed

spec/views/hq/users/index_spec.rb
theB264: There are some things in here I'm not comfortable with, Nik addressed a lot of them.  I would add to his comment though that `@users.each do` is a code fork so we should probably ensure that it correctly handles 0, 1, and more than one enumerable items in @users that all respond to `.id`, `.user_name`, `.email`, and `.active_sponsors.count`.  An array of instance_doubles of User fits perfectly.  I wouldn't check for certain tags or anything that's outside the scope of what this view does.  The data coming from `active_sponsors.count` is an example of that.  If you feed it 2 users, does it give you back 2 links to those users' show pages? How about 1? Zero @users?  That would test all of the code forks (in osra331's index view)

theB264:
includes https://osraav.atlassian.net/browse/OSRA-337

spec/views/hq/users/index_spec.rb
robbiejackson: OK, I can understand your points. I'll rework it along the lines you suggest and push up a new version for your consideration.

robbiejackson: I tried the code you suggested Nik using instance double, but it complained that "User does not implement: user_name". Funny, because I thought we had it working with that the other evening.
So I've implemented it using FactoryGirl instead and will push it up shortly.

robbiejackson: make that using a double instead of FactoryGirl - thought that aligned better with the principles Nik was expressing.

NikitaAvvakumov: I've had that problem before with `instance_double` and gave up trying to solve it. Ended up using regular `double` too.

NikitaAvvakumov:
@theB264 Feedback on the new view spec?

theB264:
:shipit:

PR #275, opened 2015-01-10
add  New Sponsor link to Sponsor show page OSRA-350

PurityControl:
https://osraav.atlassian.net/browse/OSRA-350

- given current discussions I have not added a test to prove this renders.
  I can add one if deemed important.
- Struggling to put the link to the left of "Edit Sponsor" as requested.
  Have settled to the immediate right.

### DIFF
--- a/app/admin/sponsor.rb
+++ b/app/admin/sponsor.rb
@@ -15,7 +15,6 @@ ActiveAdmin.register Sponsor do
   filter :request_fulfilled, as: :boolean
   filter :active_sponsorship_count, label: "Number of active sponsorships", as: :numeric
 
-
   actions :all, except: [:destroy]
 
   index do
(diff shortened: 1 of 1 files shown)


app/admin/sponsor.rb
```diff
@@ -148,6 +147,10 @@
   end
 
   action_item :link_to_orphan, only: :show do
+    link_to "New Sponsor", new_admin_sponsor_path
```
NikitaAvvakumov: This `action_item` needs to be renamed on L149

app/admin/sponsor.rb
```diff
@@ -147,6 +146,10 @@
     end
   end
 
+  action_item :new_orphan, only: :show do
```
theB264: Should this be `:new_orphan` or `:new_sponsor`

PurityControl: sorry - trying to do too many things at the same time, now fixed.

PR #277, opened 2015-01-11
Osra330 sponsor index views routes actions

theB264:
https://osraav.atlassian.net/browse/OSRA-330


spec/views/hq/sponsors/index_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'rails_helper'
+
+RSpec.describe "hq/sponsors/index.html.ruby", type: :view do
+  context 'sponsors exist' do
```
NikitaAvvakumov: Didn't we just merge a very different spec for users index? (#274)

app/views/hq/sponsors/index.html.ruby
```diff
@@ -0,0 +1,64 @@
...
+        </th>
+        <th>
+          Country
+        </th>
+      </tr>
+    </thead>
+    <tbody>
+    '.tap do |page|
+      @sponsors.each do |sponsor|
+        page << "
```
NikitaAvvakumov: I'll limit myself to saying that using `%Q` might be more versatile here.

theB264: This diff is outdated and another commit is on the way

theB264:
Updated

NikitaAvvakumov:
Looks ready to me. I would like to come up with something resembling consensus on spec styles so that we can honestly say things like "Follow existing code style and structure" to new team members.

spec/controllers/hq/sponsors_controller_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+end
+
+RSpec.describe Hq::SponsorsController, type: :controller do
+  before :each do
+    sign_in instance_double(AdminUser)
+    @sponsor = FactoryGirl.build_stubbed :sponsor
+  end
+
+  specify '#index' do
+    expect(Sponsor).to receive(:all).and_return(@sponsor)
```
theB264: `Model.all` should return an enumerable

theB264:
updated

app/views/hq/sponsors/index.html.haml
```diff
@@ -0,0 +1,37 @@
+- if @sponsors.empty?
```
PurityControl: I still don't see what these guards give us over the clarity that empty table provides.

app/views/hq/sponsors/index.html.haml
```diff
@@ -0,0 +1,37 @@
...
+  No Sponsors found
+- else
+  %table
```
PurityControl: do we want to make this a bootstrap table?

spec/routing/hq/sponsors_routing_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'rails_helper'
+
+RSpec.describe 'Hq::Sponsors routing', type: :routing do
```
PurityControl: I much prefer the style and readability that you used for partners and would prefer to stick to just one style.

app/views/hq/sponsors/index.html.haml
robbiejackson: I think the benefit of positively saying "no sponsors found" is that it leaves no doubt in the mind of the user. If there's an empty table there then there still may be some doubt along the lines of "is this still waiting to load the sponsor records?" or "can it not find the sponsor records?". Having said that, I output a blank users table if there were no records.

PurityControl: Given we know who are users are and their knowledge / expectations of the application, I still don't feel this is appropriate. We are only likely to get the empty set at the applications initial creation.
ActiveRecord would most likely blow up with an exception rather than return an empty set in other circumstances. It is additional logic in the view that is just not needed. In turn this creates additional branches for testing.
If we really cannot let go that we need to guard against this, although I think we should, we should push the logic back up to the controller and set a flash.

NikitaAvvakumov: @PurityControl Agree with your reasoning, but I wouldn't hold up the PR for this.

app/views/hq/sponsors/index.html.haml
NikitaAvvakumov: :+1:

app/views/hq/sponsors/index.html.haml
```diff
@@ -0,0 +1,37 @@
...
+          %td
+            = sponsor.status.name
+          %td
+            = format_full_date sponsor.start_date
+          %td
+            = sponsor.request_fulfilled ? 'Yes' : 'No'
+          %td
+            = sponsor.sponsor_type.name
+          %td
+            = ISO3166::Country.search(sponsor.country)
```
NikitaAvvakumov: Do you agree with simply adding the translated country name in this PR?

spec/routing/hq/sponsors_routing_spec.rb
NikitaAvvakumov: I'm happy with both styles, and will go with majority opinion.

spec/views/hq/sponsors/index_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
+
+RSpec.describe "hq/sponsors/index.html.haml", type: :view do
+  let(:sponsors) do
+    [ FactoryGirl.build_stubbed(:sponsor), FactoryGirl.build_stubbed(:sponsor)]
+  end
+
+  describe 'sponsors exist' do
+    before :each do
+      assign(:sponsors, sponsors)
```
NikitaAvvakumov: In view specs, since instance variables are passed to the view, there seems to be no advantage to breaking up the setup into `let` & `before` over
```ruby
before(:each) do
  @sponsors = [FactoryGirl.build_stubbed(:sponsor)]
end
```

app/views/hq/sponsors/index.html.haml
theB264: I was trying to minimize the scope, to minimize the merge time, especially on PRs that so many others are depending on.  I am all for making this a story on its own

app/views/hq/sponsors/index.html.haml
theB264: Also keep in mind that after I wrote this I noticed that User#show contains two sponsor index views, which will most likely be a partial extracted from here (probably no action items), and in that case seeing "No Sponsors Found" would be ideal

spec/routing/hq/sponsors_routing_spec.rb
theB264: I thought this way was better, if for nothing else because the method name is in the code, not in a comment.  It will appear in error messages.  But liking this way and previously receiving good feedback, I'll leave it unless one more tie-breaking vote comes in to change it to our aa-style routing specs.  @azangru @razorcd

PR #278, opened 2015-01-12
partner facade attempt

razorcd:
:8ball: 

This is the 1st attempt for writing the Partner facade. We are not happy with it because it didn't improve the controller. It added more code without simplifying the logic.

The `render`, `redirect_to`, `flash` are not accessible from the `PartnerFacade` class. 

Open for discussion. The tests should pass but ignore them for now.

Andrey and I paired on this and we disagreed a lot

### DIFF
--- a/app/controllers/hq/partners_controller.rb
+++ b/app/controllers/hq/partners_controller.rb
@@ -4,31 +4,29 @@ class Hq::PartnersController < ApplicationController
   layout 'application'
 
   def index
-    @partners = @partner_facade.load_all
+    @partners = Partner.all
   end
 
   def new
-    @partner = @partner_facade.build
-    load_associations
+    @partner_facade.build_partner
   end
   
   def create 
-    @partner = @partner_facade.build(params[:partner])
-    save_partner or re_render "new"
+    @partner_facade.build_partner(params[:partner])
+    save_partner or render :new
   end
 
   def show
-    @partner = @partner_facade.load(params[:id])
+    @partner_facade.load_partner(params[:id])
   end
 
   def edit
-    @partner = @partner_facade.load(params[:id])
-    load_associations
+    @partner_facade.load_partner(params[:id])
   end
 
   def update
-    @partner = @partner_facade.update(params[:id], params[:partner])
-    save_partner or re_render 'edit'
+    @partner_facade.update_partner(params[:id], params[:partner])
+    save_partner or render :edit
   end
 
 private
@@ -37,21 +35,10 @@ private
     @partner_facade = PartnerFacade.new
   end
 
-  def load_associations
-    @provinces = Province.all
-    @statuses = Status.all
-  end
-
   def save_partner
-    if @partner.save
+    if @partner_facade.partner.save
       flash[:success] = 'Partner successfuly saved'
-      redirect_to hq_partner_url(@partner)
+      redirect_to hq_partner_url(@partner_facade.partner)
     end
   end
-
-  def re_render(view)
-    load_associations
-    render view
-  end
-
 end
--- a/app/controllers/hq/partners_controller.rb
+++ b/app/controllers/hq/partners_controller.rb
@@ -1,49 +1,45 @@
 class Hq::PartnersController < ApplicationController
-  before_action :authenticate_admin_user!
+  before_action :authenticate_admin_user!,
+                :new_partner_facade
   layout 'application'
 
   def index
-    @partners= Partner.all
+    @partners = @partner_facade.load_all
   end
 
   def new
-    @partner = Partner.new
+    @partner = @partner_facade.build
     load_associations
   end
   
   def create 
-    build_partner
-    save_partner or re_render 'new'
+    @partner = @partner_facade.build(params[:partner])
+    save_partner or re_render "new"
   end
 
   def show
-    @partner= Partner.find(params[:id])
+    @partner = @partner_facade.load(params[:id])
   end
 
   def edit
-    load_partner
+    @partner = @partner_facade.load(params[:id])
     load_associations
   end
 
   def update
-    load_partner
-    build_partner
+    @partner = @partner_facade.update(params[:id], params[:partner])
     save_partner or re_render 'edit'
   end
 
 private
-  def load_partner
-    @partner = Partner.find(params[:id])
+
+  def new_partner_facade
+    @partner_facade = PartnerFacade.new
   end
 
   def load_associations
     @provinces = Province.all
     @statuses = Status.all
-  end
-
-  def build_partner
-    @partner ||= Partner.new
-    @partner.attributes = partner_params
   end
 
   def save_partner
@@ -56,16 +52,6 @@ private
   def re_render(view)
     load_associations
     render view
-  end
-
-  def partner_params
-    partner_params = params[:partner]
-    if partner_params
-      partner_params.permit(:name, :region, :contact_details, :province_id,
-                            :status_id, :start_date)
-    else
-      {}
-    end
   end
 
 end
--- a/app/facade/partner_facade.rb
+++ b/app/facade/partner_facade.rb
@@ -1,0 +1,38 @@
+class PartnerFacade
+
+  attr_reader :partner
+
+  def build(partner_params=nil)
+    @partner = Partner.new(permitted_params(partner_params))
+  end
+
+  def load(id)
+    @partner = Partner.find(id)
+  end
+
+  def update(id, partner_params)
+    self.load(id)
+    @partner.attributes = permitted_params(partner_params)
+    @partner
+  end
+
+  def load_all
+    Partner.all
+  end
+
+  def save
+    @partner.save
+  end
+
+private
+
+  def permitted_params(partner_params)
+    if partner_params
+      partner_params.permit(:name, :region, :contact_details, :province_id,
+                            :status_id, :start_date)
+    else
+      {}
+    end
+  end
+  
+end
--- a/app/facade/partner_facade.rb
+++ b/app/facade/partner_facade.rb
@@ -2,26 +2,25 @@ class PartnerFacade
 
   attr_reader :partner
 
-  def build(partner_params=nil)
+  def build_partner(partner_params=nil)
     @partner = Partner.new(permitted_params(partner_params))
   end
 
-  def load(id)
+  def load_partner(id)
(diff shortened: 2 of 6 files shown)


NikitaAvvakumov:
This misses the mark a bit and over-complicates our original intention. The job of a Facade is to "provide a simplified interface to a larger body of code" ([Wikipedia](http://en.wikipedia.org/wiki/Facade_pattern)). Here, we are talking about giving the views a clean unified interface to the data being passed from the controller. The facade's job is not to take over the bulk of the controller's tasks, the controller remains the key link between the view and the model. The only thing a facade - as initially discussed - should do is help the controller with data presentation. With that in mind:
- the `#index` action does not need a facade since it is already minimally simple. Adding anything complicates things needlessly.
- the facade itself tries to do too much - save partner, for example. The only job it needs to perform is create a package to be sent from the controller to the view. It should not be attempting to call `render`, `flash` etc. - those remain the controller's tasks
- `#new` & `#edit`, on the other hand, have not benefited from the facade as they still separately instantiate `@partner`, `@provinces` & `@statuses` and send that complex interface over to the view.
- the view has received no benefit at all as it still received data from the controller through multiple instance variables

The two articles that have been shared - [thoughtbot](http://robots.thoughtbot.com/sandi-metz-rules-for-developers) & [medium](https://medium.com/@ryakh/facade-pattern-on-rails-d65b86cdb5b1) - capture our original thoughts well, and I would not go beyond what they describe.

app/facade/partner_facade.rb
```diff
@@ -0,0 +1,38 @@
...
+    @partner
+  end
+
+  def load_all
+    Partner.all
+  end
+
+  def save
+    @partner.save
+  end
```
razorcd: the controller does not use the `save` from facade. This method is unused, we forgot to delete it.
Save is handled by the controller as it was before.

razorcd:
Initially I thought that we should move all logic in a facade, including render, redirect, flash, etc. So in the controller we just call, for create: `@partner.build.save` and for update: `@partner.load.update`. Chaining methods, jquery style, which also uses facade pattern as far as I know.

Thanks for the comments, I see what you mean!

azangru:
@nik: When you say that the only job a facade needs to perform is create a package to be sent from the controller to the view, do you mean that the instance variables such as `@partner`,  `@provinces` and `@statuses` should be defined in the controller (e.g. `@partner = partner_facade.partner`, `@provinces = partner_facade.provinces` etc.) and sent to the view independently or were you thinking that the whole facade package needs to be sent to the view?

Because in the first case (when instance variables are defined in the controller), the facade pattern almost does not simplify the controller, and in the second case (when the facade package is sent directly to the view), views will need to be updated.

NikitaAvvakumov:
@azangru The second case: the controller sends e.g. `@partner_facade` to the view, and the view then accesses `@partner_facade.partner`, `@partner_facade.provinces` etc. Yes, the views need to be modified. The advantage here, as I see it, is that the controller no longer has to worry about instantiating all the accessory variables; the controller/view interface is reduced to a single point; and if e.g. cities need to be sent to the view in the future, we can `def cities; City.all; end` in the facade instead of setting additional instance variables in multiple controller actions.

razorcd:
PR updated
Facade attempt 2.

Don't pay attention on the failing tests.

app/controllers/hq/partners_controller.rb
```diff
@@ -1,71 +1,44 @@
 class Hq::PartnersController < ApplicationController
-  before_action :authenticate_admin_user!
+  before_action :authenticate_admin_user!,
+                :new_partner_facade
   layout 'application'
 
   def index
-    @partners= Partner.all
+    @partners = Partner.all
   end
```
razorcd: Every controller action uses the facade except the index. Should the index use the facade too?
Like: `@partners_facade.all_partners`

razorcd:
PR updated
Facade attempt 2.

Don't pay attention on the failing tests.

app/views/hq/partners/_form.html.erb
```diff
@@ -29,7 +29,7 @@
   <div class="form-group">
     <%= f.submit class: 'btn btn-primary' %>
     <%= link_to('Cancel',
-                (@partner.id ? hq_partner_path(@partner.id) : hq_partners_path),
+                (@partner_facade.partner.id ? hq_partner_path(@partner_facade.partner) : hq_partners_path),
                 class: 'btn btn-default btn-cancel', role: 'button') %>
```
razorcd: I guess this link path logic should go inside the facade too?
And then use it like `link_to('Cancel', @partner_facade.cancel_button_path, ... )`

app/controllers/hq/partners_controller.rb
theB264: I think `@partners_facade.partners` is a good idea.  We mustn't assume we will be sending all the partners to the view.  In the three sponsors index views, there are different subsets of sponsors.  What about `load_partners :all` where `:all` is a message to the model.  So you could call `load_partners :active` if you wanted to.

app/controllers/hq/partners_controller.rb
```diff
@@ -1,71 +1,44 @@
...
   def new
-    @partner = Partner.new
-    load_associations
+    @partner_facade.build_partner
   end
   
   def create 
-    build_partner
-    save_partner or re_render 'new'
+    @partner_facade.build_partner(params[:partner])
```
theB264: `build_partner!(args) or render :new` ?  The bang could be a save as opposed to `build_partner`

spec/controllers/hq/partners_controller_spec.rb
```diff
@@ -26,20 +26,18 @@ def foobar_action
...
-    before :each do
-      controller.instance_variable_set(:@partner, partner_double)
-    end
-    
+
     specify 'successful create redirects to the show view' do
-      expect(partner_double).to receive(:save).and_return(true)
+      allow_any_instance_of(Partner).to receive(:save) do |_self|
+        _self.id = 1
+        true
```
theB264: Why true?

app/facade/partner_facade.rb
```diff
@@ -0,0 +1,37 @@
...
+
+private
+
+  def permitted_params(partner_params)
+    if partner_params
+      partner_params.permit(:name, :region, :contact_details, :province_id,
+                            :status_id, :start_date)
+    else
+      {}
```
theB264: What if you did all the params parsing here, eg
`{:id => params[:id]}.merge params.permit(:partner).permit(...)` or something to that effect
Then you could pass `params` to the facade and need no query parsing logic in the controller.

app/facade/partner_facade.rb
```diff
@@ -0,0 +1,37 @@
...
+    @partner = Partner.new(permitted_params(partner_params))
+  end
+
+  def load_partner(id)
+    @partner = Partner.find(id)
+  end
+
+  def update_partner(id, partner_params)
+    self.load_partner(id)
+    @partner.attributes = permitted_params(partner_params)
```
theB264: If you use `@partner.update_attributes= permitted(params)`, would it eliminate the need for the `save` in the controller?

app/views/hq/partners/_form.html.erb
theB264: :+1:

app/controllers/hq/partners_controller.rb
razorcd: You mean like:   `@partners_facade.load_partners :all` and `@partners_facade.load_partners :active`? This is great!

app/controllers/hq/partners_controller.rb
razorcd: Only problem is that for the a successful `build_partner!(args)` we also need to set the `flash[:successful]` message and we don't have access to that from the facade. Same with `redirect_to`.

spec/controllers/hq/partners_controller_spec.rb
razorcd: because `save` returns `true` when it successfully saves (sets an id to the current partner)
The returned `id` would also be truthy but it's not normal behavior of `save`

app/facade/partner_facade.rb
razorcd: I see what you mean but I think it's better as it is. It will be harder to read the code if we move the params parsing inside the facade and the just call:
- `@partner_facade.build_partner(params)`  instead of `@partner_facade.build_partner(params[:partner])`  **or**
- `@partner_facade.load_partner(params)` instead of `@partner_facade.load_partner(params[:id])`

Not sure. Open for discussion.

app/facade/partner_facade.rb
razorcd: That was my initial idea, to move the `save` inside the facade. But we don't have access to the `flash`, `redirect_to` and `render`. 
After discussing this in the scrum we agreed that `save` belongs in the controller.

app/controllers/hq/partners_controller.rb
theB264: Inheritance?  Include a module maybe?  If you want the flash

razorcd: tried, but didn't work for me

theB264: How about trying the reverse and include the facade as a module in the controller? Or can you fully qualify the flash somehow from ActionController::Parameters?

razorcd: I did not try that, it might work.

My concern is that by moving everything in the facade, it will make a very clean controller and a fat facade. It is not solving anything.
I think this is what Nik is also saying, the facade should only offer a "portal" for partner object manipulation and leave the flash, redirect, etc to the controller. We don't want to change rails normal behavior.  ¿

theB264: :+1:

spec/controllers/hq/partners_controller_spec.rb
theB264: Ok.  I was wondering why you needed `true` when `1` is truthy.

razorcd: you don't need `true`

theB264: `1== self.id= 1` LoL

razorcd: or `!!(_self.id= 1)` :)

PR #279, opened 2015-01-12
OSRA-348 Country names in Arabic

NikitaAvvakumov:
## Do not merge yet, waiting for client to confirm Arabic translations

https://osraav.atlassian.net/browse/OSRA-348

 - create config/locales/ar.yml with mappings of country alpha2 codes to full names in Arabic
 - in sponsor index & show views, display country names in Arabic with English in parentheses after
 - test display of English & Arabic country names in Cucumber scenario

While not directly requested by the client, this approach will allow us to demonstrate Rails' internationalization capabilities and to assure the client that separately storing data in Arabic is not necessary. Also, display of country names in both languages may help the client to catch any errors in translations before we start using the Arabic versions in orphan cards.

Complete lookup table can be seen at https://www.dropbox.com/s/n4fzgwzj7nn3o3z/CountryList%20complete%20en-ar.xlsx?dl=0


app/admin/sponsor.rb
```diff
@@ -28,8 +28,8 @@
     column :start_date
     column :request_fulfilled
     column :sponsor_type
-    column :country do |sponsor|
-      ISO3166::Country.search(sponsor.country)
+    column :country do |_sponsor|
+      "#{t _sponsor.country, locale: :ar} (#{ISO3166::Country[_sponsor.country]})"
```
theB264: With Arabic a RTL and English a LTR language, should this be formatted `(United Kingdom) المملكة المتحدة`

NikitaAvvakumov: Valid point, but both you and I are guessing. I'm OK to leave it up to the client.

PurityControl: as long as they realise they have the option.

NikitaAvvakumov:
Country now displayed as `(English) Arabic`

theB264:
:shipit:

PurityControl:
Were still waiting for confirmation - right?

NikitaAvvakumov:
Still waiting. But considering that we only really care about 4-5 countries, we can probably merge without and let them catch any mistakes after delivery.

PurityControl:
do we want @motasem-salem to check the ones that matter before we merge ?

config/locales/ar.yml
```diff
@@ -0,0 +1,225 @@
...
+  TK:
+  TO: تونغا
+  TT: ترينيداد وتوباغو
+  TN: تونس
+  TR: تركيا
+  TM: تركمانستان
+  TV: توفالو
+  UG: أوغندا
+  UA: أوكرانيا
+  AE: الإمارات العرب
```
motasem-salem: AE: الإمارات العربية المتحدة

NikitaAvvakumov:
Updated with @motasem-salem's correction

PR #280, opened 2015-01-12
OSRA353 controller authentication

theB264:
https://osraav.atlassian.net/browse/OSRA-353

### DIFF
--- a/spec/controllers/hq/hq_controller_spec.rb
+++ b/spec/controllers/hq/hq_controller_spec.rb
@@ -1,0 +1,24 @@
+require 'rails_helper'
+class NyuksController < HqController; end
+
+RSpec.describe HqController, type: :controller do
+  controller NyuksController do
+    def foobar_action
+      render nothing: true
+    end
+  end
+
+  before :each do
+    routes.draw do
+      get 'foobar_action', to: 'nyuks#foobar_action'
+    end
+  end
+
+  describe 'authentication' do
+    it 'checks for admin_user session' do
+      expect(NyuksController.superclass).to eq HqController
+      expect_any_instance_of(HqController).to receive :authenticate_admin_user!
+      get :foobar_action
+    end
+  end
+end
--- a/spec/controllers/hq/hq_controller_spec.rb
+++ b/spec/controllers/hq/hq_controller_spec.rb
@@ -1,23 +1,32 @@
 require 'rails_helper'
+
 class NyuksController < HqController; end
 
 RSpec.describe HqController, type: :controller do
+  describe 'owns namespace controllers' do
+    Rails.application.eager_load!
+    ApplicationController.descendants.each do |_controller|
+      if _controller.to_s.include? 'Hq::'
+        specify "HqController is the parent of #{_controller.to_s}" do
+          expect(_controller.superclass).to eq HqController
+        end
+      end
+    end
+  end
+
   controller NyuksController do
     def foobar_action
       render nothing: true
     end
   end
 
-  before :each do
-    routes.draw do
-      get 'foobar_action', to: 'nyuks#foobar_action'
-    end
-  end
+  describe 'controllers inheriting from HqController' do
+    specify 'check for authentication' do
+      routes.draw do
+        get 'foobar_action', to: 'nyuks#foobar_action'
+      end
+      expect_any_instance_of(Devise::Controllers::Helpers).to receive :authenticate_admin_user!
 
-  describe 'authentication' do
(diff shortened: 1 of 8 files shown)


spec/controllers/hq/users_controller_spec.rb
```diff
@@ -1,25 +1,12 @@
...
-    end
+RSpec.describe Hq::UsersController, type: :controller do
+  specify 'authentication' do
+    expect(Hq::UsersController.superclass).to eq HqController
   end
 end
-  
-describe Hq::UsersController, type: :controller do
 
+RSpec.describe Hq::UsersController, type: :controller do
```
NikitaAvvakumov: Why `RSpec` prefix?

spec/controllers/hq/partners_controller_spec.rb
```diff
@@ -1,37 +1,23 @@
...
-    end
-
-    it 'checks for admin_user session' do
-      routes.draw { get 'foobar_action', to: 'nyuks#foobar_action' }
-      expect(controller).to be_a_kind_of(Hq::PartnersController)
-      expect_any_instance_of(Hq::PartnersController).to receive :authenticate_admin_user!
-      get :foobar_action
-    end
+  specify 'authentication' do
+    expect(Hq::PartnersController.superclass).to eq HqController
```
NikitaAvvakumov: While this accomplishes the goal, I wonder if it would be better to use shared RSpec examples and `it_behaves_like` - e.g. `it_behaves_like 'HqControllerInherited'`. Not only is it more intent-revealing, but it would also exercise the child controller rather than just testing its superclass.

spec/controllers/hq/users_controller_spec.rb
theB264: If we ever turn off the Rails monkey-patching to increase app performance with `config.expose_dsl_globally= false` then this will be needed.  `RSpec.describe` does everything `describe` does, plus more, and (I think) it is more readable what is going on.

NikitaAvvakumov: Let's adopt it as app-wide syntax, then

spec/controllers/hq/hq_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  controller NyuksController do
+    def foobar_action
+      render nothing: true
+    end
+  end
+
+  describe 'anonymous child' do
+    specify 'is legitimate' do
+      expect(NyuksController.superclass).to eq HqController
```
PurityControl: What does this test do? The class is declared at the top of the test as part of the setup.

spec/controllers/hq/hq_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+  end
+
+  controller NyuksController do
+    def foobar_action
+      render nothing: true
+    end
+  end
+
+  describe 'anonymous child' do
```
PurityControl: Some sort of test description that explains we are checking for authentication or similar would be really helpful. I don't think someone new to the code base would find it easy to work out what this test is testing.

spec/controllers/hq/hq_controller_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  controller NyuksController do
+    def foobar_action
+      render nothing: true
+    end
+  end
+
+  describe 'anonymous inherited controller' do
+    specify 'checks for authentication' do
```
PurityControl: This is good enough and I'm happy for it to be merged whether changed or not. Being really picky though
`anonymous inherited controller => checks for authentication` suggest that it only checks for authentication for this controller.
I'm wondering whether it should be something like
`all controllers inheriting from hqcontroller are checked for authenticaion`

PurityControl:
I'm okay with this :arrow_up:  :ship:

NikitaAvvakumov:
Needs rebase. I also like @PurityControl's suggestion re: test description

theB264:
Rebased, updated

spec/controllers/hq/hq_controller_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    specify 'check for authentication' do
+      routes.draw do
+        get 'foobar_action', to: 'nyuks#foobar_action'
+      end
+      expect_any_instance_of(Devise::Controllers::Helpers).to receive :authenticate_admin_user!
+
+      get :foobar_action
+    end
+  end
+end
```
razorcd: :+1:

PR #282, opened 2015-01-13
osra 332 Rails edit and update for Users implemented

robbiejackson:
Changes to users controller for edit and update, implementation of associated views and tests.

### DIFF
--- a/app/controllers/hq/users_controller.rb
+++ b/app/controllers/hq/users_controller.rb
@@ -6,4 +6,26 @@ class Hq::UsersController < HqController
   def show
     @user= User.find params[:id]
   end
+
+  def edit
+    @user = User.find(params[:id])
+  end
+
+  def update
+    @user = User.find(params[:id])
+    if @user.update_attributes(user_params)
+      flash[:success] = 'User successfully saved'
+      redirect_to hq_user_url(@user)
+    else
+      render 'edit'
+    end
+  end
+
+private
+
+  def user_params
+    params.require(:user).permit(:user_name, :email)
+  end
+
+
 end
--- a/app/views/hq/users/_form.html.erb
+++ b/app/views/hq/users/_form.html.erb
@@ -1,0 +1,15 @@
+<%= form_for [:hq, @user] do |f| %>
+  <%= render 'shared/errors', object: f.object %>
+  <div class="form-group">
+    <%= f.label :user_name %>
+    <%= f.text_field :user_name, class: "form-control" %>
+  </div>
+  <div class="form-group">
+    <%= f.label :email %>
+    <%= f.text_field :email, class: "form-control" %>
+  </div>
+  <div class="form-group">
+    <%= f.submit class: "btn btn-default" %>
+    <%= link_to 'Cancel', hq_user_path(@user), class: "btn" %>
+  </div>
+<% end %>
--- a/spec/controllers/hq/users_controller_spec.rb
+++ b/spec/controllers/hq/users_controller_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe Hq::UsersController, type: :controller do
 
   before :each do
     sign_in instance_double(AdminUser)
-    @user = build_stubbed :user
+    @user = build_stubbed :user, email: 'someone@example.com'
   end
 
   it '#index' do
@@ -18,4 +18,32 @@ RSpec.describe Hq::UsersController, type: :controller do
     get :show, id: 42
     expect(response).to render_template 'show'
   end
+
+  describe '#edit and #update' do
+    before :each do
+      @old_user = FactoryGirl.build_stubbed(:user)
+      expect(User).to receive(:find).and_return(@old_user)
+    end    
+
+    specify 'editing renders the edit view' do
+      get :edit, id: @old_user.id
+      expect(response).to render_template 'edit'
+    end
+
+    it 'handles invalid updates' do
+      allow(@old_user).to receive(:save).and_return(false)
+      post :update, id: @old_user.id, user: { email: 'some email address'}
+      expect(response).to render_template 'edit'
+      expect(flash[:success]).to be_nil
+    end
+
+    it 'handles valid updates' do
+      allow(@old_user).to receive(:save).and_return(true)
+      post :update, id: @old_user.id, user: { email: 'some email address'}
+      expect(response).to redirect_to(hq_user_path(@old_user))
+      expect(flash[:success]).to_not be_nil
+    end
+
+  end
+
 end
--- a/spec/views/hq/users/_form.html.erb_spec.rb
+++ b/spec/views/hq/users/_form.html.erb_spec.rb
@@ -1,0 +1,30 @@
+require 'rails_helper'
+
+describe "hq/users/_form.html.erb", type: :view do
+
+  before :each do
+    @user = FactoryGirl.build_stubbed(:user)
+    render
+  end
+
+  it 'has a form element' do
+    expect(rendered).to have_selector('form')
+  end
+
+  it 'has an input for user name, prefilled with current value' do
+    expect(rendered).to have_selector("input\#user_user_name[value=\"#{@user.user_name}\"]")
+  end
+
+  it 'has an input for email address, prefilled with current value' do
+    expect(rendered).to have_selector("input\#user_email[value=\"#{@user.email}\"]")
+  end
+
+  it 'has a submit button' do
+    expect(rendered).to have_selector('input[type="submit"]')
+  end
+
+  it 'has a cancel button' do
+    expect(rendered).to have_link('Cancel', href: hq_user_path(@user))
+  end
+
+end
--- a/app/views/hq/users/edit.html.erb
+++ b/app/views/hq/users/edit.html.erb
@@ -1,0 +1,2 @@
+<h1> Edit User </h1>
+<%= render partial: 'form' %>
--- a/spec/views/hq/users/edit.html.erb_spec.rb
+++ b/spec/views/hq/users/edit.html.erb_spec.rb
@@ -1,0 +1,14 @@
+require 'rails_helper' 
+
+describe 'hq/users/edit.html.erb', type: :view do
+
+  before :each do
+    assign :user, build_stubbed(:user)
+  end
+
+  it 'renders the form partial' do
+    render
+    expect(view).to render_template(partial: '_form')
+  end
+
+end


spec/views/hq/users/_form.html.erb_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  it 'has an input for user name, prefilled with current value' do
+    expect(rendered).to have_selector("input\#user_user_name[value=\"#{@user.user_name}\"]")
+  end
+
+  it 'has an input for email address, prefilled with current value' do
+    expect(rendered).to have_selector("input\#user_email[value=\"#{@user.email}\"]")
+  end
+
+  it 'has a submit button' do
+    expect(rendered).to have_selector('input[type="submit"]')
```
theB264: Is this over-testing?  I'd like to get some other opinions.  I could see having tests for this if it were a client-requested feature to have certain fields maybe prefilled with certain values, but I'm not sure that's the case here.  I do like the way it's been done though.

PurityControl: I put similar tests in for partners, given that the form is an essential part of the application and is a partial that is used in a number of views. The rest of my view testing was minimal.
I don't think testing the form fields are present is too much.

robbiejackson: Well I did take inspiration from the spec for the equivalent form in the Partners area, which I think is testing the same thing. Except that it uses assert_select and I felt I didn't fully understand what way it was working.

spec/controllers/hq/users_controller_spec.rb
```diff
@@ -31,4 +31,35 @@ def foobar_action
...
+      new_user = controller.instance_variable_get(:@user)
+      expect(new_user[:email]).to eq 'some email address'
+      expect(flash[:success]).to be_nil
+    end
+
+    it 'handles valid updates' do
+      allow(@old_user).to receive(:save).and_return(true)
+      post :update, id: @old_user.id, user: { email: 'some email address'}
+      expect(response).to redirect_to(hq_user_path(@old_user))
+      new_user = controller.instance_variable_get(:@user)
```
robbiejackson: On second thoughts I think I should remove lines 58-60. If it's going to redirect to show, then it doesn't matter what the @user instance variable is set to.

app/views/hq/users/_form.html.erb
```diff
@@ -0,0 +1,15 @@
...
+    <%= f.label :user_name %>
+    <%= f.text_field :user_name, class: "form-control" %>
+  </div>
+  <div class="form-group">
+    <%= f.label :email %>
+    <%= f.text_field :email, class: "form-control" %>
+  </div>
+  <div class="form-group">
+    <%= f.submit class: "btn btn-default" %>
+    <%= link_to 'Cancel', hq_user_path(@user), class: "btn" %>
```
NikitaAvvakumov: `btn-cancel`?

app/controllers/hq/users_controller.rb
```diff
@@ -6,4 +6,24 @@ def index
...
+      redirect_to hq_user_url(@user)
+    else
+      render 'edit'
+    end
+  end
+
+private
+
+  def user_params
+    params[:user].permit(:user_name, :email)
```
NikitaAvvakumov: Why not also use `require`?

app/controllers/hq/users_controller.rb
```diff
@@ -6,4 +6,24 @@ def index
     @users = User.all
   end
 
+  def edit
+    @user = User.find(params[:id])
```
NikitaAvvakumov: Unless there are well-founded objections, let's stick to the example set by Partners controller with `load_user` etc.?

spec/controllers/hq/users_controller_spec.rb
```diff
@@ -31,4 +31,33 @@ def foobar_action
     expect(response).to render_template 'index'
   end
 
+  describe '#edit and #update' do
```
NikitaAvvakumov: Again, the Partners controller and related specs have established a generally agreed-upon paradigm. Is there a particular reason for the differences here?

spec/views/hq/users/_form.html.erb_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'rails_helper'
+
+describe "hq/users/_form.html.erb", type: :view do
+
+  before :each do
+    @user = FactoryGirl.build_stubbed(:user, user_name: 'Frederick & Bloggs', email: 'fred.bloggs@example.com')
```
NikitaAvvakumov: No need to set `user_name` & `email` explicitly

app/views/hq/users/_form.html.erb
robbiejackson: it's in the next pull request

app/controllers/hq/users_controller.rb
robbiejackson: Personally I found the Partners code more difficult to follow. When there's a simple statement like on line 10, why complicate it by replacing it with a function call, forcing you to go down to it to follow the code?

theB264: I am also a fan of leaving the single line here, unless the complexity grows or the code is repeated, then definitely abstracting it

PurityControl: I prefer the function call I'm afraid.

app/controllers/hq/users_controller.rb
robbiejackson: Why include require? Won't the model catch any problems where the user hasn't specified the mandatory fields? The Partners controller didn't have a require call.

NikitaAvvakumov: Including `require` shortcuts the process so that it fails before going to the model. Additionally, the failure is clearer: the model will report that a required attribute is missing, but it won't be immediately obvious why. With `required`, the controller will throw `ActionController::ParameterMissing`, making it clear that the form failed to submit a required parameter.

If you look at #252, you'll see that there was some discussion and disagreement about this and a related issue. The Partners controller does not use `require` because of a certain aspect of its architecture that the Users controller does not replicate.

robbiejackson: Nik I'm really not sure what you're asking me to do about avoiding the model in the case where a mandatory parameter is missing. Are you suggesting that I don't attempt a save in that case and add manually the error message for the user that a mandatory field is missing? This means that any errors on any other fields won't be highlighted, where it's the model which does the validation (eg email address format). 
In the case that the user enters one mandatory field (eg username) but not, say, email address, how do I get the mandatory field missing error message set against the email field but not the username field?

NikitaAvvakumov: The only change I'm suggesting is
```ruby
def user_params
  params.require(:user).permit(:user_name, :email)
end
```
If the params hash is missing the `user` key, the controller will automatically raise the `ParameterMissing` error without even attempting to create or update a user record.

robbiejackson: Oh ok, sorry I thought you meant to require the mandatory attributes user_name and email.
So what, if anything, should I do about handling the exception?

NikitaAvvakumov: No need to do anything. It would be impossible to trigger this exception if the app is built and tested correctly, and if the end user operates via the UI. If either of those isn't the case, it's appropriate for the app to fail loudly.

app/controllers/hq/users_controller.rb
theB264: See next comment :arrow_down:

app/controllers/hq/users_controller.rb
```diff
@@ -6,4 +6,24 @@ def index
...
+  end
+
+  def update
+    @user = User.find(params[:id])
+    if @user.update_attributes(user_params)
```
theB264: If you extract out `User.find` you could combine lines 14-15 here `if method.update_attr....`

app/views/hq/users/_form.html.erb
theB264: This will probably raise an exception if the form is used for `new` as `@user.id` will be `nil`.  Suggest a conditional link destination

PurityControl: or just link back to the index view

NikitaAvvakumov: @robbiejackson: the issue raised by @theB264 needs to be addressed.

spec/views/hq/users/_form.html.erb_spec.rb
NikitaAvvakumov: I would like to see this cleaned up since the values of `user_name` & `email` are not used explicitly in the test.

spec/controllers/hq/users_controller_spec.rb
```diff
@@ -31,4 +31,33 @@ def foobar_action
...
+      get :edit, id: @old_user.id
+      expect(response).to render_template 'edit'
+    end
+
+    it 'handles invalid updates' do
+      allow(@old_user).to receive(:save).and_return(false)
+      post :update, id: @old_user.id, user: { email: 'some email address'}
+      expect(response).to render_template 'edit'
+      new_user = controller.instance_variable_get(:@user)
+      expect(new_user[:email]).to eq 'some email address'
```
NikitaAvvakumov: L49 & 50 are testing that the newly rendered 'edit' form uses the object's new attributes carried from the original form submission. Assignment of the new attributes to the `@user` object is done by Rails as part of the call to `.update_attributes` and not by our controller. This test is, therefore, unnecessary.

NikitaAvvakumov:
@robbiejackson I see 3 points that need to be addressed, and the branch needs to be rebased prior to merge.

spec/views/hq/users/_form.html.erb_spec.rb
robbiejackson: Yes ok, will remove the explicity setting of user_name and email. I had set a specific value of the user name which included an ampersand to check if I needed to do anything wrt html escaping (which is done in the equivalent partners spec). But having confirmed that it's not necessary there's no longer any need to set the fields explicitly.

spec/controllers/hq/users_controller_spec.rb
robbiejackson: Well really update_attributes sets the new attributes correctly only if it's the @user object that you're calling it on. So really this test is checking that, for example, you haven't left the @ off the @user accidentally in your controller code. However, no matter, I'll remove the lines.

NikitaAvvakumov:
Superseded by #283

PR #283, opened 2015-01-13
Osra333 users new and create

robbiejackson:
Users controller changed to support #new and #create.
Associated views and test specs created.

### DIFF
--- a/app/controllers/hq/users_controller.rb
+++ b/app/controllers/hq/users_controller.rb
@@ -21,6 +21,21 @@ class Hq::UsersController < HqController
     end
   end
 
+  def new
+    @user = User.new
+  end
+
+  def create
+    @user = User.new
+    @user.attributes = user_params
+    if @user.save
+      flash[:success] = 'User successfully created'
+      redirect_to hq_user_url(@user)
+    else
+      render 'new'
+    end
+  end
+
 private
 
   def user_params
--- a/app/controllers/hq/users_controller.rb
+++ b/app/controllers/hq/users_controller.rb
@@ -6,4 +6,25 @@ class Hq::UsersController < HqController
   def show
     @user= User.find params[:id]
   end
+
+  def edit
+    @user = User.find(params[:id])
+  end
+
+  def update
+    @user = User.find(params[:id])
+    if @user.update_attributes(user_params)
+      flash[:success] = 'User successfully saved'
+      redirect_to hq_user_url(@user)
+    else
+      render 'edit'
+    end
+  end
+
+private
+
+  def user_params
+    params.require(:user).permit(:user_name, :email)
+  end
+
 end
--- a/spec/controllers/hq/users_controller_spec.rb
+++ b/spec/controllers/hq/users_controller_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe Hq::UsersController, type: :controller do
 
   before :each do
     sign_in instance_double(AdminUser)
-    @user = build_stubbed :user, email: 'someone@example.com'
+    @user = build_stubbed :user
   end
 
   it '#index' do
@@ -34,8 +34,6 @@ RSpec.describe Hq::UsersController, type: :controller do
       allow(@old_user).to receive(:save).and_return(false)
       post :update, id: @old_user.id, user: { email: 'some email address'}
       expect(response).to render_template 'edit'
-      new_user = controller.instance_variable_get(:@user)
-      expect(new_user[:email]).to eq 'some email address'
       expect(flash[:success]).to be_nil
     end
 
@@ -44,6 +42,43 @@ RSpec.describe Hq::UsersController, type: :controller do
       post :update, id: @old_user.id, user: { email: 'some email address'}
       expect(response).to redirect_to(hq_user_path(@old_user))
       expect(flash[:success]).to_not be_nil
+    end
+
+  end
+
+  describe '#new' do
+    
+    before :each do
+      @new_user = User.new
+      allow(User).to receive(:new).and_return(@new_user)
+    end
+
+    it 'enables entering details of a new user' do
+      get :new
+      expect(response).to render_template 'new'
+      expect(assigns(:user)).to be_a_new(User)
+    end
+
+  end
+
+  describe '#create' do
+
+    before :each do
+      @new_user = FactoryGirl.build_stubbed(:user, user_name: nil, email: nil)
+      allow(User).to receive(:new).and_return(@new_user)
+    end
+
+    specify 'successful create shows the new user' do
+      expect(@new_user).to receive(:save).and_return(true)
+      post :create, user: { user_name: 'some user' }
+      expect(response).to redirect_to hq_user_path(@new_user)
+      expect(flash[:success]).to_not be_nil
+    end
+
+    specify 'unsuccessful create re-renders the new view' do
+      expect(@new_user).to receive(:save).and_return(false)
+      post :create, user: { user_name: 'some user' }
+      expect(response).to render_template 'new'
     end
 
   end
--- a/spec/controllers/hq/users_controller_spec.rb
+++ b/spec/controllers/hq/users_controller_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe Hq::UsersController, type: :controller do
 
   before :each do
     sign_in instance_double(AdminUser)
-    @user = build_stubbed :user
+    @user = build_stubbed :user, email: 'someone@example.com'
   end
 
   it '#index' do
@@ -18,4 +18,34 @@ RSpec.describe Hq::UsersController, type: :controller do
     get :show, id: 42
     expect(response).to render_template 'show'
   end
+
+  describe '#edit and #update' do
+    before :each do
+      @old_user = FactoryGirl.build_stubbed(:user)
+      expect(User).to receive(:find).and_return(@old_user)
+    end    
+
+    specify 'editing renders the edit view' do
+      get :edit, id: @old_user.id
+      expect(response).to render_template 'edit'
+    end
+
+    it 'handles invalid updates' do
+      allow(@old_user).to receive(:save).and_return(false)
+      post :update, id: @old_user.id, user: { email: 'some email address'}
+      expect(response).to render_template 'edit'
+      new_user = controller.instance_variable_get(:@user)
+      expect(new_user[:email]).to eq 'some email address'
+      expect(flash[:success]).to be_nil
+    end
+
+    it 'handles valid updates' do
+      allow(@old_user).to receive(:save).and_return(true)
+      post :update, id: @old_user.id, user: { email: 'some email address'}
+      expect(response).to redirect_to(hq_user_path(@old_user))
+      expect(flash[:success]).to_not be_nil
+    end
+
+  end
+
 end
--- a/spec/views/hq/users/_form.html.erb_spec.rb
+++ b/spec/views/hq/users/_form.html.erb_spec.rb
@@ -1,0 +1,30 @@
+require 'rails_helper'
+
+describe "hq/users/_form.html.erb", type: :view do
+
+  before :each do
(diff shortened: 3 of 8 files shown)


spec/controllers/hq/users_controller_spec.rb
```diff
@@ -31,4 +31,74 @@ def foobar_action
...
+      expect(response).to redirect_to(hq_user_path(@old_user))
+      expect(flash[:success]).to_not be_nil
+    end
+
+  end
+
+  describe '#new' do
+    
+    before :each do
+      @new_user = FactoryGirl.build_stubbed(:user, user_name: nil, email: nil)
```
PurityControl: Why factory girl for a new user?

spec/controllers/hq/users_controller_spec.rb
```diff
@@ -31,4 +31,74 @@ def foobar_action
...
+      allow(User).to receive(:new).and_return(@new_user)
+    end
+
+    it 'enables entering details of a new user' do
+      get :new
+      expect(response).to render_template 'new'
```
PurityControl: this seems more convoluted than just accessing the @new_user variable in the before block

spec/controllers/hq/users_controller_spec.rb
```diff
@@ -31,4 +31,74 @@ def foobar_action
...
+      user = controller.instance_variable_get(:@user)
+      expect(user).to_not be_nil
+      expect(user[:user_name]).to be_nil
```
PurityControl: I'm not sure what this expect buys us. It may be nil for multiple reasons but I can't ever see it being magically not nil.

spec/controllers/hq/users_controller_spec.rb
```diff
@@ -31,4 +31,74 @@ def foobar_action
...
+    specify 'editing renders the edit view' do
+      get :edit, id: @old_user.id
+      expect(response).to render_template 'edit'
+    end
+
+    it 'handles invalid updates' do
+      allow(@old_user).to receive(:save).and_return(false)
+      post :update, id: @old_user.id, user: { email: 'some email address'}
+      expect(response).to render_template 'edit'
+      new_user = controller.instance_variable_get(:@user)
```
PurityControl: You could check this using assigns. But either way I think this leads to brittle testing.

spec/controllers/hq/users_controller_spec.rb
```diff
@@ -31,4 +31,74 @@ def foobar_action
...
+      post :create, user: { user_name: 'some user' }
+      expect(response).to redirect_to hq_user_path(@new_user)
+      expect(flash[:success]).to_not be_nil
+    end
+
+    specify 'unsuccessful create re-renders the new view' do
+      expect(@new_user).to receive(:save).and_return(false)
+      post :create, user: { user_name: 'some user' }
+      expect(response).to render_template 'new'
+      user = controller.instance_variable_get(:@user)
```
PurityControl: as above, not keen on checking variable assignment in a controller

spec/controllers/hq/users_controller_spec.rb
robbiejackson: So just use User.new instead, yes?

spec/controllers/hq/users_controller_spec.rb
robbiejackson: I've convinced myself that checking that the controller has set the instance variables correctly is important, as it's one of the key requirements of the controller. Partly this is due to the fact that I didn't initially have checks for that, and my tests were all passing, but the functionality wasn't working correctly due to a bug in handling the params.

spec/controllers/hq/users_controller_spec.rb
robbiejackson: So I'm suggesting instead of lines 73-75
expect(assigns(:user)).to be_a_new(User)

spec/controllers/hq/users_controller_spec.rb
robbiejackson: Proposed change is below, in the next comment

spec/controllers/hq/users_controller_spec.rb
PurityControl: yes please

spec/controllers/hq/users_controller_spec.rb
PurityControl: perhaps a discussion for a scrum then, and a more convincing argument for ubiquitous use of facades as testing could be pushed to the facade and leave just a single assigns test in the controller.
However, I would prefer to leave these out and have a simple functional test that checks that records in the db are displayed in the view

spec/controllers/hq/users_controller_spec.rb
robbiejackson: What I was trying to check was that the email parameter sent in the post request actually got set in that entry in the @users variable, so that it appeared when the edit form was re-rendered. How do I use assigns to get at an individual entry in the @users hash? (Again, this was something which wasn't working initially because of the way I was doing the update).

spec/controllers/hq/users_controller_spec.rb
PurityControl: that would work, bearing in mind, the comments I have already made on assigns vs functional tests.

spec/controllers/hq/users_controller_spec.rb
robbiejackson: yes, let's discuss on the scrum. I'll fix up the other bits and push them up.

spec/controllers/hq/users_controller_spec.rb
PurityControl: Minitest doesn't separate out controllers and views in the way rspec does. It seems the test you are trying to check for sits in no mans land as the controllers and view testing gets split in rspec.
Something to add to our scrum discussion perhaps.

spec/controllers/hq/users_controller_spec.rb
theB264: I would think that checking internal use of variable assignments is bad, but in the case where a controller is setting vars to be passed to the view in a closure, this is a key function a controller serves, because it's the controller's way of returning data in rails, and thus needs testing.  I would think that adding any tests for a view or that data is displayed is not good.

app/controllers/hq/users_controller.rb
```diff
@@ -6,4 +6,39 @@ def index
...
+      redirect_to hq_user_url(@user)
+    else
+      render 'new'
+    end
+  end
+
+private
+
+  def user_params
+    params[:user].permit(:user_name, :email)
```
theB264: Would this buy us anything here? `params.require(:user).permit(:user_name, :email)`

spec/controllers/hq/users_controller_spec.rb
theB264: If this is what you intend to do, you may be able to `expect(assigns(:user).email).to eq 'string'`.  I see why you want to do this but I'm on the fence as to whether we should.  I'm okay either way

spec/controllers/hq/users_controller_spec.rb
```diff
@@ -31,4 +31,72 @@ def foobar_action
...
+      expect(response).to render_template 'new'
+      expect(assigns(:user)).to be_a_new(User)
+    end
+
+  end
+
+  describe '#create' do
+
+    before :each do
+      @new_user = FactoryGirl.build_stubbed(:user, user_name: nil, email: nil)
```
theB264: Using the real class for a new, unsaved instance shouldn't hit the db.  @new_user= User.new

spec/controllers/hq/users_controller_spec.rb
theB264: `expect(assigns(:inst_var).method_name).to eq object` or similar if we're going to check this

spec/views/hq/users/new.html.erb_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'rails_helper' 
+
+describe 'hq/users/new.html.erb', type: :view do
+
+  before :each do
+    assign :user, build_stubbed(:user)
```
theB264: If you use `User.new` here it shouldn't have an `:id`, more similar to the object that would come from the controller

app/controllers/hq/users_controller.rb
NikitaAvvakumov: :+1: `params.require(...).permit(...)` is a useful Rails convention. Prefer to use it unless there is good reason against it. Partners controller has been updated to this style as well.

spec/controllers/hq/users_controller_spec.rb
NikitaAvvakumov: See my comment on #282: the test is not necessary

spec/controllers/hq/users_controller_spec.rb
```diff
@@ -22,7 +22,7 @@ def foobar_action
 
   before :each do
     sign_in instance_double(AdminUser)
-    @user = build_stubbed :user
+    @user = build_stubbed :user, email: 'someone@example.com'
```
NikitaAvvakumov: Again, not sure why the explicit assignment of the `email` attribute

spec/controllers/hq/users_controller_spec.rb
NikitaAvvakumov: @robbiejackson If your intention is to check that the params are handled correctly, then that needs to become a separate test that thoroughly goes through the params hash and ensures that all permitted params are accepted and others are rejected. e.g. see https://gist.github.com/jeremyw/3805223. If you decide to go this route, please submit it as a separate PR.

I remain of the opinion that these two lines do not belong in this test (please see my other comments here and in #282).

spec/views/hq/users/_form.html.erb_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'rails_helper'
+
+describe "hq/users/_form.html.erb", type: :view do
+
+  before :each do
+    @user = FactoryGirl.build_stubbed(:user, user_name: 'Frederick & Bloggs', email: 'fred.bloggs@example.com')
```
NikitaAvvakumov: As previously

NikitaAvvakumov:
Essentially the same few issues as with #282 & rebase and it's ready to go.

app/controllers/hq/users_controller.rb
robbiejackson: Agreed, will fix params.require

spec/controllers/hq/users_controller_spec.rb
robbiejackson: removed as requested (hang-over from edit/update pull request)

spec/controllers/hq/users_controller_spec.rb
robbiejackson: No I can't think why I would have explicitly assigned the email attribute. Will rmove this.

spec/controllers/hq/users_controller_spec.rb
robbiejackson: OK, I'll remove these lines checking the instance variables are set correctly. However, I remain unconvinced. If the user enters invalid data in a field and hits submit then when the form is re-rendered he should see in the field the invalid data he previously entered rather than, say, a blank field, and enabling this via correct setting of the instance variable is what these 2 lines are testing. It's not to do with accepting / rejecting params.

spec/views/hq/users/_form.html.erb_spec.rb
robbiejackson: Fixed (hang-over from edit/update)

spec/controllers/hq/users_controller_spec.rb
NikitaAvvakumov: Setting of the instance variable is covered by the tests indirectly but thoroughly: 
- you tell the User model to return the factory-generated object when it receives a call to `find`;
- you then tell the factory object to expect a `save` message;
- and finally you verify than on successful save there is a redirection to the factory object's page

All of the above expectations can only be met if the instance variable is set and used correctly throughout the controller action. Explicitly testing that this is so, however, means testing the implementation, which is unnecessarily low-level and brittle.

PR #284, opened 2015-01-13
OSRA354 add cancel btn

theB264:
https://osraav.atlassian.net/browse/OSRA-354

### DIFF
--- a/app/views/hq/partners/_form.html.erb
+++ b/app/views/hq/partners/_form.html.erb
@@ -2,35 +2,35 @@
   <%= render 'shared/errors', object: f.object %>
   <div class="form-group">
     <%= f.label :name %>
-    <%= f.text_field :name, class: 'form-control' %>
+    <%= f.text_field :name, class: "form-control" %>
   </div>
   <div class="form-group">
     <%= f.label :region %>
-    <%= f.text_field :region, class: 'form-control' %>
+    <%= f.text_field :region, class: "form-control" %>
   </div>
   <div class="form-group">
     <%= f.label :contact_details %>
-    <%= f.text_field :contact_details, class: 'form-control' %>
+    <%= f.text_field :contact_details, class: "form-control" %>
   </div>
   <div class="form-group">
     <%= f.label :province_id %>
     <%= f.collection_select :province_id,
-     @provinces, :id, :name , {},  class: 'form-control' %>
+     @provinces, :id, :name , {},  class: "form-control" %>
   </div>
   <div class="form-group">
     <%= f.label :status_id %>
     <%= f.collection_select :status_id,
-      @statuses, :id, :name, {}, class: 'form-control' %>
+      @statuses, :id, :name, {}, class: "form-control" %>
   </div>
   <div class="form-group">
     <%= f.label :start_date %>
-    <%= f.text_field :start_date, class: 'form-control date-picker' %>
+    <%= f.text_field :start_date, class: "form-control date-picker" %>
   </div>
   <div class="form-group">
     <%= f.submit class: 'btn btn-primary' %>
     <%= link_to('Cancel',
                 (@partner.id ? hq_partner_path(@partner.id) : hq_partners_path),
-                class: 'btn btn-default', role: 'button') %>
+                class: 'btn btn-default btn-cancel', role: 'button') %>
   </div>
 <% end %>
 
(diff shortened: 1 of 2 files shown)


theB264:
screenshot from 2015-01-13 17 51 34

app/views/hq/partners/_form.html.erb
```diff
@@ -2,32 +2,35 @@
...
     <%= f.label :start_date %>
-    <%= f.text_field :start_date, class: "form-control date-picker" %>
+    <%= f.text_field :start_date, class: 'form-control date-picker' %>
   </div>
   <div class="form-group">
-    <%= f.submit class: "btn btn-default" %>
+    <%= f.submit class: 'btn btn-primary' %>
+    <%= link_to('Cancel',
+                (@partner.id ? hq_partner_path(@partner.id) : hq_partners_path),
+                class: 'btn btn-default', role: 'button') %>
```
PurityControl: I thought we had agreed in the scrum that every one would use the semantic 'btn-cancel'

spec/views/hq/partners/_form.html.erb_spec.rb
```diff
@@ -1,25 +1,39 @@
...
-    assert_select "form"
+    render and assert_select 'form'
   end
 
-  specify 'renders each form value' do
+  describe '"Cancel" button' do
+    specify 'when no :id' do
+      allow(partner).to receive(:id).and_return nil
+      render
+      assert_select 'a[href=?]', hq_partners_path, role: 'button', text: 'Cancel'
```
PurityControl: I think checking for where a link goes and possibly (?) its wording is more than enough. Tying styling in to the test makes it too brittle.

spec/views/hq/partners/_form.html.erb_spec.rb
```diff
@@ -1,25 +1,39 @@
...
+    end
+
+    specify 'when :id' do
+      allow(partner).to receive(:id).and_return 42
+      render
+      assert_select 'a[href=?]', hq_partner_path(42), role: 'button', text: 'Cancel'
```
PurityControl: as above

spec/views/hq/partners/_form.html.erb_spec.rb
```diff
@@ -1,25 +1,39 @@
...
+    end
+  end
+
+  specify 'form values' do
+    render
```
PurityControl: I like the line between the render and the tests. I'm not bothered which style we use but can we pick one and use it consistently.

app/views/hq/partners/_form.html.erb
```diff
@@ -2,32 +2,35 @@
...
   <div class="form-group">
     <%= f.label :start_date %>
-    <%= f.text_field :start_date, class: "form-control date-picker" %>
+    <%= f.text_field :start_date, class: 'form-control date-picker' %>
   </div>
   <div class="form-group">
-    <%= f.submit class: "btn btn-default" %>
+    <%= f.submit class: 'btn btn-primary' %>
+    <%= link_to('Cancel',
+                (@partner.id ? hq_partner_path(@partner.id) : hq_partners_path),
```
robbiejackson: In OSRA 333 for Users I implemented the selection of different cancel paths in a different way - basically by passing the cancel path to use in as a locals parameter to the form partial. I think it would make sense to adopt a common approach across the project. Which way do you guys prefer?

theB264: In general, I prefer a simpler approach unless the complexity buys us something.

spec/views/hq/partners/_form.html.erb_spec.rb
theB264: :arrow_up:  outdated diff

app/views/hq/partners/_form.html.erb
```diff
@@ -2,32 +2,35 @@
   <%= render 'shared/errors', object: f.object %>
   <div class="form-group">
     <%= f.label :name %>
-    <%= f.text_field :name, class: "form-control" %>
+    <%= f.text_field :name, class: 'form-control' %>
```
PurityControl: if we have agreed to allow people to submit code in their preferred style can we not change previously submitted code.

PurityControl:
I'm okay with this.

app/views/hq/partners/_form.html.erb
NikitaAvvakumov: @theB264 Agreed

PR #285, opened 2015-01-14
OSRA 335 form labels changed to add an asterisk if field is mandatory

robbiejackson:
form_builder label method changed to determine if the field passed is mandatory (ie presence: true in the model), and append an asterisk if that's the case.

### DIFF
--- a/config/initializers/form_builder.rb
+++ b/config/initializers/form_builder.rb
@@ -1,0 +1,22 @@
+class ActionView::Helpers::FormBuilder
+  alias :orig_label :label
+
+# add a '*' after the field label if the field is required
+# see http://davidsulc.com/blog/2011/05/01/self-marking-required-fields-in-rails-3/
+# Also http://blog.pothoven.net/2012/10/self-marking-required-fields-in-rails.html
+  def label(method, content_or_options = nil, options = nil, &block)
+    if content_or_options && content_or_options.class == Hash
+      options = content_or_options
+    else
+      content = content_or_options
+    end
+
+    required_mark = ''
+    required_mark = ' *'.html_safe if object.class.validators_on(method).map(&:class).include? ActiveRecord::Validations::PresenceValidator
+
+    content ||= I18n.t("activerecord.attributes.#{object.class.name.underscore}.#{method}", :default=>method.to_s.humanize)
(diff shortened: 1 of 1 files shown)


config/initializers/form_builder.rb
```diff
@@ -0,0 +1,22 @@
+class ActionView::Helpers::FormBuilder
```
PurityControl: @NikitaAvvakumov @theB264 :arrow_up: 
https://3.bp.blogspot.com/-a0a5yxl0Uhw/ThXNmvrV6mI/AAAAAAAAAc0/ZvL-qUccbR8/s1600/2595_250.gif

PurityControl:
@robbiejackson if you scream in some code with no tests, can anyone hear it? :sob:

config/initializers/form_builder.rb
```diff
@@ -0,0 +1,22 @@
...
+# Also http://blog.pothoven.net/2012/10/self-marking-required-fields-in-rails.html
+  def label(method, content_or_options = nil, options = nil, &block)
+    if content_or_options && content_or_options.class == Hash
+      options = content_or_options
+    else
+      content = content_or_options
+    end
+
+    required_mark = ''
+    required_mark = ' *'.html_safe if object.class.validators_on(method).map(&:class).include? ActiveRecord::Validations::PresenceValidator
```
robbiejackson: This doesn't totally work for Partners. In particular for the Province and Status fields, but the form controls for those fields don't allow a blank setting anyway (at least not in the hq implementation). 
The label functionality works by checking if the field passed in has got a validator presence: true in the model.
For Province it doesn't work because the form field is province_id, and in the model it has
validates :province, presence: true
For Status, the model doesn't have any associated presence: true.
One solution could be to set a CSS class of "required" against a field, instead of adding the asterisk, and then use CSS to add the asterisk. With that way you could set the CSS "required" class manually in the form against anything else like Province or Status. 
Thoughts?

robbiejackson:
Yes, I should have said that I was only pushing it up to get some feedback. It causes lots of the cucumber tests to fail.

NikitaAvvakumov:
Decided on alternative approach in team meeting

PR #287, opened 2015-01-15
OSRA-351 Display full country names in filter

NikitaAvvakumov:
https://osraav.atlassian.net/browse/OSRA-351

Original filter only displayed 2-letter country codes that are stored with the sponsor record. The new implementation provides the filter with a hash that maps those codes to full country names.


NikitaAvvakumov:
Rebased & green

PR #261, opened 2015-01-04
Osra315 hq cucumber tests for authentication

razorcd:
https://osraav.atlassian.net/browse/OSRA-315

**Notice:**
- Model specs are run together with the aa spec tests!
- Cucumber tests are running in verbose mode: it lists all the loaded files when starts
- features/support is shared between the cucumber:aa and cucumber:hq tests
- features/support/step_helper.rb has the helper methods for hq step definitions 
(features/support folder is always loaded first by cucumber)

**Commands:**
- `$rake`    # -> will run all tests: spec(aa and hq), `cucumber:hq` and `cucumber:aa`
- `$rake test:aa` # -> will run all aa tests: `spec:aa`, `cucumber:aa`
- `$rake test:hq` # -> will run all hq tests: `spec:hq`, `cucumber:hq`
- `$rake spec:aa` # -> will run all aa specs: all specs that are not in a 'hq' folder
- `$rake spec:hq` # -> will run all hq specs: all specs that are in a 'hq' folder
- `$rake cucumber:aa` # -> will run all aa cucumber tests: all features from features/aa and will load all .rb files from  features/aa and features/support
- `$rake cucumber:hq` # -> will run all hq cucumber tests: all features from features/hq and will load all .rb files from  features/hq and features/support

- `$cucumber` and `$cucumber -p aa` # -> will run features in 'aa' profile
- `$cucumber -p hq` # -> will run features in 'hq' profile
- `$rspec spec/controllers/hq/users_controller_spec.rb` $ -> will run rspec tests from one file
- `$cucumber -p aa features/aa/aa_features/users.feature` # -> will run one feature file in 'aa' profile
- `$cucumber -p hq features/hq/hq_features/authentication.feature` # -> will run one feature file in 'hq' profile

PR #289, opened 2015-01-15
OSRA-356: Partner's province shouldn't be editable

motasem-salem:
https://osraav.atlassian.net/browse/OSRA-356

### DIFF
--- a/app/views/hq/partners/_form.html.erb
+++ b/app/views/hq/partners/_form.html.erb
@@ -15,7 +15,7 @@
   <div class="form-group">
     <%= f.label :province_id %>
     <%= f.collection_select :province_id,
-     @provinces, :id, :name , {},  class: "form-control" %>
+     @provinces, :id, :name , {},  class: "form-control", disabled: !f.object.new_record?%>
   </div>
   <div class="form-group">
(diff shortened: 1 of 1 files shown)


theB264:
Should we put a test for this in the spec or is this considered eye-candy because the model rejects updating this? related: https://github.com/AgileVentures/osra/pull/213

motasem-salem:
In AA, we tested this in a feature. I assumed it will be the same in this case.

theB264:
I'm also struggling trying to get a feature written before https://github.com/AgileVentures/osra/pull/261 is merged.

PurityControl:
I'm okay with this

NikitaAvvakumov:
Do you think a view spec would be more appropriate here? We didn't have the option of writing those for AA but can for HQ. A feature seems to be overkill to test the behaviour of logic within a single view. I'd prefer to limit features to high-level exercises that ensure that multiple parts of the app function together.

PurityControl:
Given this is a flourish rather than a functionality issue I'm not sure it warrants a test. We probably wouldn't put this in a test if it was part of the original feature.

theB264:
I'm okay either way

PR #291, opened 2015-01-16
OSRA-363 OSRA-364 Default translation values

NikitaAvvakumov:
https://osraav.atlassian.net/browse/OSRA-363
https://osraav.atlassian.net/browse/OSRA-364

AA: Provide global default of _[Translation N/A]_ to handle cases where Arabic translation is missing in Sponsor index & show and User show views. Prefer this to an empty string to make such cases easily identifiable for possible fixing.
screen shot 2015-01-17 at 11 20 21 pm

screen shot 2015-01-17 at 11 55 46 pm

### DIFF
--- a/app/admin/sponsor.rb
+++ b/app/admin/sponsor.rb
@@ -32,7 +32,7 @@ ActiveAdmin.register Sponsor do
     column :request_fulfilled
     column :sponsor_type
     column :country do |_sponsor|
-      "(#{ISO3166::Country[_sponsor.country]}) #{t _sponsor.country, locale: :ar}"
+      en_ar_country(_sponsor.country)
     end
   end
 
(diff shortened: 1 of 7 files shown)


app/admin/sponsor.rb
```diff
@@ -32,7 +32,7 @@
     column :request_fulfilled
     column :sponsor_type
     column :country do |_sponsor|
-      "(#{ISO3166::Country[_sponsor.country]}) #{t _sponsor.country, locale: :ar}"
+      "(#{ISO3166::Country[_sponsor.country]}) #{t _sponsor.country, locale: :ar, default: ''}"
```
PurityControl: Is this something than can and if it can, should be set globally?

PurityControl: http://guides.rubyonrails.org/i18n.html, Section 4.1.2 says we can use a dictionary value as a default which would allow us to only change the message in one place if it had to change.

Quote from above:

If the :default value is a Symbol, it will be used as a key and translated. One can provide multiple values as default. The first one that results in a value will be returned.

E.g., the following first tries to translate the key :missing and then the key :also_missing. As both do not yield a result, the string "Not here" will be returned:

```
I18n.t :missing, default: [:also_missing, 'Not here']
# => 'Not here'
```

Thoughts?

NikitaAvvakumov: That's why you should keep reading the docs even after you've found the answer :) :+1:

PR #292, opened 2015-01-16
osra355 setting asterisk on mandatory fields

robbiejackson:
Some other possibilities for setting * against mandatory fields.
1. (Partner :name field) Set attribute “require” on input field. Use javascript to add the * onto the associated label. (This has been implemented for <input> elements only and would need to be extended for other types of input controls). 
2. (Partner :region field) Set "required-field" class on the label which is associated with the mandatory input field.
3. (Partner :contact_details field) Set "required-group" on the form-group which is a parent of the mandatory input field.
Note that this has just been pushed up for comments, not for merging.

### DIFF
--- a/app/assets/javascripts/require-star.js
+++ b/app/assets/javascripts/require-star.js
@@ -1,0 +1,3 @@
+$(document).ready(function(){
+  $("input[required]").parent().children("label").append("<span class=\"mandatory-field\">*</span>");
+});
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -13,3 +13,4 @@
  *= require_self
  *= require_tree .
  */
+@import "requirestar"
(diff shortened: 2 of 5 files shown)


NikitaAvvakumov:
@robbiejackson I like the solution from http://blog.pothoven.net/2012/10/self-marking-required-fields-in-rails.html, which is the only one that would prevent things getting out of sync between the model validations and the views. I would prefer the original approach taken in the blog post of setting a CSS class on required labels rather than appending a asterisk directly, as this would give us more flexibility. I also wonder whether this solution would noticeably slow down form rendering since each label will have to additionally consult the model, though I suspect that the answer is "No".

robbiejackson:
@nik That was the previous solution I proposed - see pull request 285 (https://github.com/AgileVentures/osra/pull/285) but we agreed not to go down that route for the reasons stated in that pull request.

NikitaAvvakumov:
Thanks for reminding me! It's a shame that that solution did not work for all our use cases.

NikitaAvvakumov:
Will reopen when @robbiejackson returns or if someone takes over the task

PR #293, opened 2015-01-17
refactor partners controller to use params.require

PurityControl:
Given build_partner is not used for its original intended purpose and the conversations about params.require in #282.
#partner_params has now been refactored to remove the unnecessary guards and introduce params.require in its place.


theB264:
:shipit:

PR #294, opened 2015-01-17
OSRA-349 sponsorship destruction

NikitaAvvakumov:
https://osraav.atlassian.net/browse/OSRA-349

As per client request, enable sponsorship destruction, to be used in cases of
incorrectly created sponsorships. `after_destroy` callback on the Sponsorship
model instructs pertinent Sponsor & Orphan records to update themselves. The
update functionality for Sponsor & Orphan is awkwardly repurposed from already
existing - will refactor (see below).

-----
### Original PR description

Notes:
- Sponsor "reset" upon sponsorship destruction is the exact same callback that takes place on sponsorship save
- Orphan "reset" is the same callback that takes place on orphan reactivation (status change from "On Hold" to "Active")

At the very least, in the case of Orphan I would like to unify the disjointed actions that take place on sponsorship creation, sponsorship inactivation, sponsorship destruction, and orphan reactivation into a single interface. Ideally, however, I would like to experiment with extracting a service object that handles all necessary events whenever sponsorship changes state.

Given that the functionality is now in place, but very much cobbled together out of existing pieces rather than designed with any degree of forethought or elegance, what are people's thoughts on how to proceed with the refactoring I propose (which might become substantial)? If the decision is to merge, then I will definitely proceed with the refactoring ASAP instead of leaving it for "later".

### DIFF
--- a/app/admin/sponsor.rb
+++ b/app/admin/sponsor.rb
@@ -89,6 +89,10 @@ ActiveAdmin.register Sponsor do
             input '', type: :hidden, name: :authenticity_token, value: form_authenticity_token
           text_node '</form>'.html_safe
         end
+        column '' do |_sponsorship|
+          link_to 'X', admin_sponsor_sponsorship_path(sponsor_id: sponsor.id, id: _sponsorship.id), method: :delete,
+            data: { confirm: 'WARNING: You are about to permanently delete the record of this sponsorship!' }
+        end
       end
     end
 
--- a/app/models/orphan.rb
+++ b/app/models/orphan.rb
@@ -124,6 +124,10 @@ class Orphan < ActiveRecord::Base
     current_sponsorship.sponsor if currently_sponsored?
   end
 
+  def resolve_sponsorship_status
+    reactivate and save!
+  end
+
 private
 
   def sponsored_siblings_does_not_exceed_siblings_count
--- a/lib/tasks/clear_orphans.rake
+++ b/lib/tasks/clear_orphans.rake
@@ -5,7 +5,7 @@ namespace :db do
     OrphanList.destroy_all
     PendingOrphan.destroy_all
     PendingOrphanList.destroy_all
-    Sponsorship.destroy_all
+    Sponsorship.delete_all # skip callbacks on already destroyed orphan records
 
     Sponsor.update_all(:request_fulfilled => false)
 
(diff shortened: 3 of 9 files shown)


app/admin/sponsor.rb
```diff
@@ -89,6 +89,10 @@
             input '', type: :hidden, name: :authenticity_token, value: form_authenticity_token
           text_node '</form>'.html_safe
         end
+        column '' do |_sponsorship|
+          link_to 'X', admin_sponsor_sponsorship_path(sponsor_id: sponsor.id, id: _sponsorship.id), method: :delete,
```
NikitaAvvakumov: Ideas on how to make this less accessible and/or scarier are welcome

app/models/orphan.rb
```diff
@@ -124,6 +124,10 @@ def current_sponsor
     current_sponsorship.sponsor if currently_sponsored?
   end
 
+  def resolve_sponsorship_status
+    reactivate and save!
```
NikitaAvvakumov: `#reactivate` has all the necessary functionality, but its name is obviously ill-suited here

lib/tasks/clear_orphans.rake
```diff
@@ -5,7 +5,7 @@ namespace :db do
     OrphanList.destroy_all
     PendingOrphan.destroy_all
     PendingOrphanList.destroy_all
-    Sponsorship.destroy_all
+    Sponsorship.delete_all
```
NikitaAvvakumov: Bypass callbacks, otherwise Sponsorship tries to call `resolve_sponsorship_status` on orphan that has been deleted already.

app/admin/sponsor.rb
```diff
@@ -89,6 +89,10 @@
...
+            data: { confirm: 'WARNING: You are about to permanently delete the record of this sponsorship!' }
```
theB264: Might we add an Arabic translation here?

theB264:
screenshot from 2015-01-18 14 12 24

lib/tasks/clear_orphans.rake
PurityControl: I'm normally against comments but perhaps in this case a comment making intent clear might be helpful.

PurityControl:
Given this is a client request, the current state of technical debt and the fact that you are reusing what's there rather than adding to it, Merging this in wouldn't greatly upset me.
I agree with your strategy for refactoring.
I think we should aim for a series of smaller refactorings rather than a substantial one.

PR #295, opened 2015-01-18
OSRA-369 Make sponsor agent mandatory

NikitaAvvakumov:
https://osraav.atlassian.net/browse/OSRA-369

Assignment of agent to sponsor is now mandatory as per client request.
Only new test is in sponsor model spec.
Factories & features modified to pass.

**All sponsors currently on production have agents assigned.** I have asked the client to continue doing so for all new sponsors. Once this PR is merged, I'll add agents to all sponsors from the CLI on testing, and then will do same after pushing to staging.

### DIFF
--- a/app/models/sponsor.rb
+++ b/app/models/sponsor.rb
@@ -29,6 +29,7 @@ class Sponsor < ActiveRecord::Base
   validate :can_be_inactivated, if: :being_inactivated?, on: :update
   validates_format_of :email, with: Devise.email_regexp, allow_blank: true
   validate :type_matches_affiliation, on: :create
+  validates :agent, presence: true
 
   belongs_to :branch
   belongs_to :organization
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -1,5 +1,5 @@
 FactoryGirl.define do
-  factory :user do
+  factory :user, aliases: [:agent] do
     sequence(:user_name) { |n| "#{Faker::Name.name} #{n}" }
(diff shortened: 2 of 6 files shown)


spec/factories/users.rb
```diff
@@ -1,5 +1,5 @@
 FactoryGirl.define do
-  factory :user do
+  factory :user, aliases: [:agent] do
```
theB264: What is this doing?

NikitaAvvakumov: Allows to refer to the factory as `agent`. e.g. in sponsors factory and in the sponsor feature.

spec/models/sponsor_spec.rb
```diff
@@ -47,6 +47,7 @@
   it { is_expected.to belong_to :sponsor_type }
   it { is_expected.to have_many(:orphans).through :sponsorships }
   it { is_expected.to belong_to :agent }
+  it { is_expected.to validate_presence_of :agent }
```
motasem-salem: Here as well

app/models/sponsor.rb
```diff
@@ -37,6 +37,7 @@ class Sponsor < ActiveRecord::Base
   has_many :sponsorships
   has_many :orphans, through: :sponsorships
   belongs_to :agent, :class_name => 'User', :foreign_key => 'agent_id'
+  validates :agent, presence: true
```
motasem-salem: Maybe move this line up to be with the other validates lines.

NikitaAvvakumov:
Updated & passing

PR #296, opened 2015-01-18
OSRA329 user#show

theB264:
https://osraav.atlassian.net/browse/OSRA-329


theB264:
screenshot from 2015-01-18 08 31 07

theB264:
screenshot from 2015-01-18 08 33 48

app/views/hq/sponsors/_index.html.haml
```diff
@@ -0,0 +1,38 @@
...
+            %td
+              = sponsor.status.name
+            %td
+              = format_full_date sponsor.start_date
+            %td
+              = sponsor.request_fulfilled ? 'Yes' : 'No'
+            %td
+              = sponsor.sponsor_type.name
+            %td
+              = "(#{ISO3166::Country[sponsor.country]}) #{t sponsor.country, locale: :ar}"
```
NikitaAvvakumov: Can now use the `en_ar_country(sponsor.country)` helper here

app/views/hq/sponsors/index.html.haml
```diff
@@ -1,38 +1 @@
...
-              = sponsor.status.name
-            %td
-              = format_full_date sponsor.start_date
-            %td
-              = sponsor.request_fulfilled ? 'Yes' : 'No'
-            %td
-              = sponsor.sponsor_type.name
-            %td
-              = "(#{ISO3166::Country[sponsor.country]}) #{t sponsor.country, locale: :ar}"
+= render partial: 'hq/sponsors/index.html.haml', locals: {sponsors: @sponsors}
```
NikitaAvvakumov: Don't know how well this will work with HAML, but can try `= render @sponsors`
http://guides.rubyonrails.org/layouts_and_rendering.html#using-partials, section 3.4.5 Rendering Collections

spec/views/hq/users/show_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+        expect(rendered).to match user.user_name
+      end
+
+      specify 'email address' do
+        expect(rendered).to match user.email
+      end
+    end
+  end
+
+  [['active', 'Active'], ['active', 'On Hold'], ['inactive', 'Inactive']].each do |status|
```
NikitaAvvakumov: So... let's change this?

app/views/hq/sponsors/index.html.haml
theB264: no-go

spec/views/hq/users/show_spec.rb
theB264: View partial spec needed.  In the works

app/views/hq/sponsors/index.html.haml
PurityControl: I would guess it less to do with templating engines and more to do with namespacing issues

theB264: Yep, it looks for hq/sponsors/_sponsors.html.*

theB264:
The last commit (9ac03f5) changes the "empty" partials to a nicer, matching format and forces both renders or neither, but not one only.  Of note that this is an eye-candy commit and the tests were not changed, they were written to pass either way.

screenshot from 2015-01-20 00 26 07

app/views/hq/sponsors/_index.html.haml
```diff
@@ -0,0 +1,43 @@
+- if sponsors.empty?
+  .table-responsive
+    %table.table
+      %tbody
+        %tr
+          %td
+            No Sponsors found
```
PurityControl: should a No Sponsors Found message really be put in a table ?

app/views/hq/sponsors/_index.html.haml
```diff
@@ -0,0 +1,43 @@
+- if sponsors.empty?
```
PurityControl: what is a sponors index partial doing in a user#show story?

theB264: user#show has two sponsor indexes, that's why it waited on the osra330 merge.  This just reformats the existing view to be reusable by User

app/views/hq/sponsors/_index.html.haml
theB264: Much easier to match the styling between the two when they're the same thing.  I think we still need eye candy stories for most of this anyway, this just moves it from terrible to acceptable before those get tackled.  So, yes, I think it's a good idea at this point.  In the finished product, defintely not.  But right now I'm happy if the funcitonality is added while eye candies are pending

PurityControl: Given there is no tabulated data to show, something like a panel would be much more appropriate.
http://getbootstrap.com/components/#panels

PR #302, opened 2015-01-19
[WIP] OSRA-326 Partner Facade

NikitaAvvakumov:
# DO NOT MERGE
## for comparison & discussion only. cf. #278

https://osraav.atlassian.net/browse/OSRA-326

This is my idea of what a facade should do: package data for presentation to
the view. In the partners controller, its utility is negligible since most
actions do not instantiate multiple variables and their interface to the views
is already minimal. With that in mind, having explored the use of a facade in
the partners controller as per our initial intent, I am not certain that its
advantages outweigh the costs here. Other controllers, however, may well benefit from
facades. Sponsor show & User show both display the main object's associations,
and these associations feel like ideal candidates for loading from a facade.

In #278, the facade takes on multiple responsibilities and thus violates
SOLID guidelines. In addition to sending data from the controller to the view,
it also gets involved in building and updating the partner object with
attributes returned from the view to the controller. The facade also takes
over validation of params, which in my mind clearly falls into the controller's
area of responsibility.

### DIFF
--- a/app/controllers/hq/partners_controller.rb
+++ b/app/controllers/hq/partners_controller.rb
@@ -7,13 +7,12 @@ class Hq::PartnersController < ApplicationController
   end
 
   def new
-    @partner = Partner.new
-    load_associations
+    @facade = PartnerFacade.new(Partner.new)
   end
 
   def create
     build_partner
-    save_partner or re_render 'new'
+    save_partner or render 'new'
   end
 
   def show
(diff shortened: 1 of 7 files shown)


app/controllers/hq/partners_controller.rb
```diff
@@ -21,14 +20,13 @@ def show
...
-    load_partner
-    load_associations
+    @facade = PartnerFacade.new(load_partner)
   end
 
   def update
     load_partner
     build_partner
-    save_partner or re_render 'edit'
+    save_partner or render 'edit'
```
PurityControl: perhaps load the facade in `load_partner` or have a `load_facade` method to keep things dry, as a failed save in update will require a handle to the `@facade` instance variable

razorcd:
:+1: 
I see what you mean. And as you say, the partners controller is too simple for a facade to be helpful.
This is why I felt it's better to move all the logic from the controller to the facade. I guess it depends how you see the use of the facade because the way I made it would break some of the rails conventions.

PR #303, opened 2015-01-20
changes for performance improvements on render of orphan index page OSRA-366

PurityControl:
https://osraav.atlassian.net/browse/OSRA-366
Eager loading and conditional initialization showing ~3x speed improvement
on initial page render time on develop.
- remove duplicates from city of origin filter and sort alphabetically
- guard to only execute initializers on a new record
- add default scope

### DIFF
--- a/app/models/orphan.rb
+++ b/app/models/orphan.rb
@@ -66,6 +66,8 @@ class Orphan < ActiveRecord::Base
   accepts_nested_attributes_for :current_address, allow_destroy: true
   accepts_nested_attributes_for :original_address, allow_destroy: true
 
+  default_scope { includes(:partner, :orphan_sponsorship_status, :orphan_status, original_address: :province) }
+
   def father_name
     "#{father_given_name} #{family_name}"
   end
@@ -134,11 +136,15 @@ private
   end
 
   def default_sponsorship_status_unsponsored
-    self.orphan_sponsorship_status ||= OrphanSponsorshipStatus.find_by_name 'Unsponsored'
+    if new_record?
+      self.orphan_sponsorship_status ||= OrphanSponsorshipStatus.find_by_name 'Unsponsored'
+    end
   end
 
   def default_orphan_status_active
-     self.orphan_status ||= OrphanStatus.find_by_name 'Active'
+    if new_record?
+      self.orphan_status ||= OrphanStatus.find_by_name 'Active'
+    end
   end
 
   def valid_date? date
--- a/app/admin/orphan.rb
+++ b/app/admin/orphan.rb
@@ -5,8 +5,8 @@ ActiveAdmin.register Orphan do
   filter :gender, as: :select, collection: Settings.lookup.gender
   filter :province_code, as: :select,
          collection: proc { Province.distinct.map { |p| [p.name, p.code] } }
-  filter :original_address, label: 'City of origin', as: :select,
-         collection: proc { Address.original.map { |a| [a.city, a.id] } }
+  filter :original_address_city, label: 'City of origin', as: :select,
+        collection: proc {Orphan.distinct.pluck(:city).sort.map{|c| [c, c]} }
   filter :priority, as: :select
   filter :orphan_sponsorship_status, as: :select,
          collection: proc { OrphanSponsorshipStatus.all.map { |oss| [oss.name, oss.id] } }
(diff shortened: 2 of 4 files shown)


theB264:
I'm okay with the code you changed.  Some of the code that's there is ripe for refactor, but outside the scope of this PR.  :shipit:

app/models/orphan.rb
```diff
@@ -66,6 +66,8 @@ class Orphan < ActiveRecord::Base
   accepts_nested_attributes_for :current_address, allow_destroy: true
   accepts_nested_attributes_for :original_address, allow_destroy: true
 
+  default_scope { includes(:partner, :orphan_sponsorship_status, :orphan_status, original_address: :province) }
```
NikitaAvvakumov: Is this too much eager loading for the default scope? Every `find` etc. query will load up 5 AR objects in addition to the orphan itself. Should this be limited to the orphans controller?

Somewhat related - switching statuses from associations to `enums` will boost performance in addition to other benefits. Is it time yet?

PurityControl: I don't think so, especially for our current situation.
- For a find that returns a single record the difference will be largely unnoticeable, even if the relationships weren't needed.
- For returning large row sets in the index view, the benefit gained by eager loading is dramatic.
- If there were any situations where the relationships weren't needed (I can't think of any off hand  where we wouldn't use something like pluck) we could override the default scope.
- Active admin uses the default scope for filters so for production as it stands at the moment I think it is the best choice for performant rendering.

If this irritates us post active admin I think we could consider other options, but for the moment I think it is our best choice. 
 
As for enums, the last time we talked about this we were holding off until active admin was out of the picture so we didn't have to amend code we were going to be letting go of. I think this is probably still the best course of action.

PurityControl: To further clarify, I understand the concern, and had initially started with a narrower scope. The active admin filtering issue made this bubble up to the default scope.
Given this, I think the points in the post above hold true.

NikitaAvvakumov: Would this work for filters: http://activeadmin.info/docs/2-resource-customization.html#customizing_resource_retrieval?

PurityControl: No, it doesn't work for filters.

app/models/orphan.rb
```diff
@@ -134,11 +136,15 @@ def sponsored_siblings_does_not_exceed_siblings_count
...
+    if new_record?
+      self.orphan_sponsorship_status ||= OrphanSponsorshipStatus.find_by_name 'Unsponsored'
+    end
   end
 
   def default_orphan_status_active
-     self.orphan_status ||= OrphanStatus.find_by_name 'Active'
+    if new_record?
+      self.orphan_status ||= OrphanStatus.find_by_name 'Active'
+    end
```
NikitaAvvakumov: There is also `default_priority_to_normal` on L158. What are your thoughts on combining all these initializers into a single method?

PurityControl: I wouldn't be upset if the codes was presented to me like that. But changing from what we have seems to fly in the face of a method having a single responsibility.

NikitaAvvakumov: One could argue that the single responsibility is initializing an instance in that case

PurityControl: Agreed, but we already have a method that does that, `after_initialize`.
To do as you are proposing we could override `after_initialize` as a method, and merge the code from the three methods into one single method, but I think that loses clarity compatred to the method call at the top of the class we currently have.
But to create another method that has the same responsibility that `after_initialize` has, that `after_initialize` calls, I don't see value in.

NikitaAvvakumov: I was thinking that this would be cleaner:
```ruby
after_initialize :set_defaults

def set_defaults
  return unless new_record?
  self.orphan_sponsorship_status ||= ...
  self.orphan_status ||= ...
  self.priority ||= ...
end
```
No need to override anything. `after_initialize` currently calls 3 methods that set defaults for a new instance, instead it could be calling one.

This is a minor point anyway.

NikitaAvvakumov:
@PurityControl Can you please add a `if new_record?` clause to `default_priority_to_normal` on L158 of app/models/orphan.rb?

NikitaAvvakumov:
Superseded by #321

PR #304, opened 2015-01-20
Osra336 sponsor#edit sponsor#update actions and views

razorcd:
https://osraav.atlassian.net/browse/OSRA-336

Created:
 - controller edit#sponsor
 - view edit#sponsor
 - controller update#sponsor
 - partial view _form
 - `country_select_list` and `en_ar_country(country)` view helper methods to hq: `application_helper.rb`
 - rspec tests for controller and views
 - changed `:sponsor` factory to include `address`, `email`, `contact1`, `contact2`, `additional_info`
 
Comments will be removed on the next update.


app/controllers/hq/sponsors_controller.rb
```diff
@@ -5,4 +5,56 @@ class Hq::SponsorsController < ApplicationController
...
+    render view
+  end
+
+  def sponsor_params   #for #create we don't permit :sponsor_type_id, :organization_id, :branch_id. security issue?!?
+    params.require(:sponsor)
+        .permit(:name, :address, :country, :email, :contact1, :contact2,
+                :additional_info, :status_id, :start_date, :sponsor_type_id,
+                :gender, :branch_id, :organization_id, :osra_num, :sequential_id,
+                :requested_orphan_count, :agent_id, :city, :new_city_name, :payment_plan)
+  end
```
razorcd: for #create, which is a different task, we don't permit `:sponsor_type_id`, `:organization_id`, `:branch_id`.
Should we have a different `sponsor_params` for #create?

app/helpers/application_helper.rb
```diff
@@ -9,4 +11,36 @@ def format_full_date(date)
...
   end
+
+  def en_ar_country(country)
+    en_country = ISO3166::Country[country]
+    ar_country = I18n.with_locale :ar do
+      t country, default: :missing
+    end
+
+    ("(#{en_country}) #{ar_country}").html_safe
+  end
```
razorcd: this is the same method from `admin`

app/views/hq/sponsors/_form.html.haml
```diff
@@ -0,0 +1,101 @@
...
+    = f.label :agent
+    = f.select :agent_id, User.pluck(:user_name, :id), 
+          {include_blank: true}, class: "form-control"
+ 
+  %div.form-group
+    = f.submit class: 'btn btn-primary'
+    = link_to 'Cancel',
+                (@sponsor.id ? hq_sponsors_path : hq_sponsors_path),
+                class: 'btn btn-default btn-cancel', role: 'button'
+    /(@sponsor.id ? hq_sponsor_path(@sponsor.id) : hq_sponsors_path)
```
razorcd: Cance links will be changed once sponsor#show action will be merged.

spec/factories/sponsors.rb
```diff
@@ -16,6 +16,11 @@
...
     organization { Organization.all.sample if sponsor_type.name == 'Organization' }
     payment_plan { Sponsor::PAYMENT_PLANS.sample }
+    address { "#{Faker::Address.street_name} #{Faker::Address.building_number.to_s}" }
+    email { Faker::Internet.email }
+    contact1 { Faker::PhoneNumber.phone_number }
+    contact2 { Faker::PhoneNumber.cell_phone }
+    additional_info { Faker::Lorem.sentence }
     agent
   end
 end
```
razorcd: I added new fields to the `:sponsor` factory to check their presence in the `_form` partial. These new fields are all optional in the model (no presence validation).
Should I just create a new `:full_sponsor` factory instead of editing `:sponsor`? (I think not)

spec/views/hq/sponsors/edit_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+RSpec.describe 'hq/sponsors/edit.html.haml', type: :view do
+  before :each do
+    assign :sponsor, build_stubbed(:sponsor)
+    assign :statuses, []
+    assign :sponsor_types, []
+    assign :organizations, []
+    assign :branches, []
+    assign :cities, []
+  end
```
razorcd: I added `[ ]` instead of ex: `Status.all` just to speed up things. I am checking for the presence of these lists in the `_form` partial anyway.

app/controllers/hq/sponsors_controller.rb
theB264: `hash.reject{|k, _| [:sym1, :sym2].include? k }`

app/views/hq/sponsors/_form.html.haml
theB264: It was previously decided early-on that we would include the dead links to the unfinished `hq` actions.

spec/factories/sponsors.rb
theB264: :+1: :+1: :+1:  We need this in more factories.  I haven't done it because there was pushback before for adding non-mandatory info, but +3 for that idea.  I needed this yesterday in my PR

spec/factories/sponsors.rb
```diff
@@ -16,6 +16,11 @@
     branch { Branch.all.sample if sponsor_type.name == 'Individual' }
     organization { Organization.all.sample if sponsor_type.name == 'Organization' }
     payment_plan { Sponsor::PAYMENT_PLANS.sample }
+    address { "#{Faker::Address.street_name} #{Faker::Address.building_number.to_s}" }
```
theB264: for non-mandatory fields
``` ruby
attribute [nil, your_code].sample
```
may be even better at catching errors

razorcd: clever! good point, I'll add it like that

spec/factories/sponsors.rb
razorcd: :+1:

theB264: If this change survives code review, you'll be seeing a story for adding these in all the factories, so...

PurityControl: Vanilla factories should really only have mandatory fields. If you want additional fields I think traits would be a better choice.
https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#traits

app/controllers/hq/sponsors_controller.rb
```diff
@@ -5,4 +5,56 @@ class Hq::SponsorsController < ApplicationController
   def index
     @sponsors= Sponsor.all
   end
+
+  def edit
+    # @sponsor= Sponsor.new   #for testing #new
```
PurityControl: can we remove the comment ?

app/helpers/application_helper.rb
```diff
@@ -1,6 +1,8 @@
 module ApplicationHelper
   FULL_DATE_FORMAT_STRING= '%d %B %Y'
   MONTH_YEAR_DATE_FORMAT_STRING= '%m/%Y'
+  PRIORITY_COUNTRIES= %w(SA TR AE GB)
```
PurityControl: I would prefer to see these in the sponsor model for the time being

app/helpers/application_helper.rb
```diff
@@ -9,4 +11,36 @@ def format_full_date(date)
...
+  def en_ar_country(country)
+    en_country = ISO3166::Country[country]
+    ar_country = I18n.with_locale :ar do
+      t country, default: :missing
+    end
+
+    ("(#{en_country}) #{ar_country}").html_safe
+  end
+
+  def country_select_list
```
PurityControl: I don't think this is needed.
https://github.com/stefanpenner/country_select#usage

spec/controllers/hq/sponsors_controller_spec.rb
```diff
@@ -31,4 +31,26 @@ def foobar_action
     expect(response).to render_template 'index'
   end
 
+  specify '#edit' do
```
PurityControl: :+1: You seem to be hitting a nice sweet spot for controller testing

app/helpers/application_helper.rb
razorcd: Yes. There is a gem for everything! :( 
It doesn't feel like real programming. It feels more like just playing with lego.

razorcd:
sponsor edit

razorcd:
PR Updated (and rebased)

app/models/sponsor.rb
```diff
@@ -2,6 +2,8 @@ class Sponsor < ActiveRecord::Base
   include Initializer
   NEW_CITY_MENU_OPTION = '**Add New**'
   PAYMENT_PLANS = ['Monthly', 'Every Two Months', 'Every Four Months', 'Every Six Months', 'Annually', 'Other']
+  PRIORITY_COUNTRIES= %w(SA TR AE GB)
+  EXCLUDED_COUNTRYS= %w(IL BQ BV IO KY CX CC CW FK TF GG HM IM JE YT AN NF NO PN BL SH MF PM SX GS SS SJ TK TC UM AX)
```
razorcd: All these countries does not have translations so I excluded them.

spec/factories/sponsors.rb
```diff
@@ -17,5 +17,42 @@
...
+    trait :random_optional_fields do  
+      [true,false].sample and address_street
+      [true,false].sample and email
+      [true,false].sample and contact1
+      [true,false].sample and contact2
+      [true,false].sample and additional_info
+    end
+
+    factory :sponsor_full, traits: [:address_street, :email, :contact1, :contact2, :additional_info]
+    factory :sponsor_random_optional_fields, traits: [:random_optional_fields]
```
razorcd: I tried to also create a `:sponsor_random_optional_fields` factory that has the optional fields displayed or NOT(empty). Randomly.
Only problem with this is that the fields randomness only occurs once you reload `FactoryGirl` (on `$rspec` start or on a `FactoryGirl.reload`).

Randomness will not occur when you call 
`FactoryGirl.build(:sponsor_random_optional_fields)` and then you call again 
`FactoryGirl.build(:sponsor_random_optional_fields)`.
The same fields will be `nil` or `not nil`

I played with this for hours but no better solution.
I think it is still useful like this.
What do you think?

app/models/sponsor.rb
theB264: So, if we add translations, we also need to edit this constant?

app/views/hq/sponsors/_form.html.haml
```diff
@@ -0,0 +1,104 @@
...
+
+  %div.form-group
+    = f.label :email
+    = f.text_field :email, {class: "form-control", type: "email"}
+
+  %div.form-group
+    = f.label :contact1
+    = f.text_field :contact1, class: "form-control"
+
+  %div.form-group
```
theB264: `%div` is unneeded

app/models/sponsor.rb
razorcd: YES.
Or we can change `en_ar_country` to return only the English version on a missing translation.
But the missing countries are not that important:

BQ (Bonaire, Sint Eustatius and Saba) 
BV (Bouvet Island)
IO (British Indian Ocean Territory)
KY (Cayman Islands)
CX (Christmas Island)
CC (Cocos (Keeling) Islands)
CW (Curaçao)
FK (Falkland Islands (Malvinas)) 
TF (French Southern Territories)
GG (Guernsey)
HM (Heard and McDonald Islands)
IM (Isle of Man) 
JE (Jersey)
YT (Mayotte)
AN (Netherlands Antilles)
NF (Norfolk Island) 
PN (Pitcairn)
BL (Saint Barthélemy) 
SH (Saint Helena)
MF (Saint Martin) 
PM (Saint Pierre And Miquelon)
SX (Sint Maarten) 
GS (South Georgia and the South Sandwich Islands) 
SS (South Sudan)
SJ (Svalbard And Jan Mayen)
TK (Tokelau)
TC (Turks and Caicos Islands) 
UM (United States Minor Outlying Islands) 
AX (Åland Islands)

theB264: I personally prefer the former option

razorcd: not sure why "NO" : "Norway"  is not translating.
`I18n.t 'NO'` => `"translation missing: ar.NO"`
and we have NO defined in ar.yml

theB264: To clarify (I misread your comment) the option I prefer is
"change `en_ar_country` to return only the English version on a missing translation."

spec/factories/sponsors.rb
PurityControl: have you tried something along the lines off....
```
trait :random_optional_fields do  
  address_street { [true,false,nil].sample }
  email { [true,false,nil].sample } 
  contact1 { [true,false,nil].sample }
  contact2 { [true,false,nil].sample }
  additional_info { [true,false,nil].sample }
end
```

razorcd:
PR Updated. All fixed.

app/helpers/admin/country_helper.rb
```diff
@@ -3,7 +3,8 @@ module Admin::CountryHelper
   def en_ar_country(country)
     en_country = ISO3166::Country[country]
     ar_country = I18n.with_locale :ar do
-      t country, default: :missing
+        country = "NOR" if country == "NO" #yml "NO" bug
```
theB264: Instead of this workaround, can we use `'NO': النرويج` instead of `NO: النرويج` in the yml file?  I think it works...

PurityControl: @razorcd can you try this before we merge this

app/views/hq/sponsors/_form.html.haml
```diff
@@ -0,0 +1,105 @@
...
+    = f.label :requested_orphan_count
+    = f.text_field :requested_orphan_count, class: "form-control"
+  - unless f.object.new_record?
+    .form-group
+      = f.label :request_fulfilled
+      = f.check_box :request_fulfilled, {disabled: true, class: "checkbox"},
+           f.object.request_fulfilled ? 1 : 0
+
+
+  - if f.object.new_record?
```
PurityControl: this is where a facade would help - you could have a method #class_for_attribute rather than a conditional for the view.
Without a facade i'm tempted to put this into a view helper

spec/views/hq/sponsors/_form_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+        assert_select "option", value: sponsor_random_optional_fields.sponsor_type_id,
+                                html: CGI::escape_html(sponsor_random_optional_fields.sponsor_type.name) do
+            assert_select "[selected=?]", "selected"
+        end
+      end
+
+      assert_select "select#sponsor_organization_id" do
+        assert_select "[disabled=?]", "disabled"
+        assert_select "option", value: sponsor_random_optional_fields.organization_id
+        assert_select "option", html: CGI::escape_html(sponsor_random_optional_fields.organization.name) if sponsor_random_optional_fields.organization
```
PurityControl: we tend to prefer an 80 char line width and only use a trailing if, when it can fit into these constraints.
perhaps a smaller variable name such as `full_sponsor` or `sponsor_with_opts`.

app/views/hq/sponsors/_form.html.haml
```diff
@@ -0,0 +1,105 @@
...
+  .form-group
+    = f.label :gender
+    = f.select :gender, Settings.lookup.gender, {}, class: "form-control"
+  .form-group
+    = f.label :start_date
+    = f.text_field :start_date, class: "form-control date-picker"
+  .form-group
+    = f.label :requested_orphan_count
+    = f.text_field :requested_orphan_count, class: "form-control"
+  - unless f.object.new_record?
```
theB264: Would it be better to put this block inside the `else` below?

app/helpers/admin/country_helper.rb
razorcd: sure, will try this evening

spec/views/hq/sponsors/_form_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+      ["name", "requested_orphan_count", "start_date", "new_city_name", "address", "email", 
+       "contact1", "contact2", "additional_info"].each do |field|
+        assert_select "input#sponsor_#{field}" do
+          if sponsor_full[field]
+            assert_select "[value=?]", CGI::escape_html(sponsor_full[field].to_s) 
+          else
+            assert_select "[value]", false
+          end
+        end
+      end
```
razorcd: I made this to remove the repetitive tests. 
I am not a fan of this way because it is hard to read but I gave it a try. I can change it back if requested.
What do you think?

razorcd: This is the error that is shows on a wrong `sponsor.name` field:

```
hq/sponsors/_form.html.haml
  has a form
  has a "Cancel" button
    using an existing Sponsor record
    using a new Sponsor record
  has form values
    using an existing Sponsor record (FAILED - 1)
    using a new Sponsor record

Failures:

  1) hq/sponsors/_form.html.haml has form values using an existing Sponsor record
     Failure/Error: assert_select "[value=?]", CGI::escape_html(sponsor_full[field].to_s)
     Minitest::Assertion:
       Expected at least 1 element matching "[value='Mr. Bean']", found 0.
       Expected 0 to be >= 1.
```

spec/views/hq/sponsors/_form_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+
+      assert_select "input#sponsor_request_fulfilled", false
+
+      #disabled selects
+      ["sponsor_type", "organization", "branch"].each do |field|
+        assert_select "select#sponsor_#{field}_id" do
+          assert_select "[disabled]", false
+        end
+      end
```
razorcd: same here as above

razorcd:
PR Updated

config/locales/ar.yml
```diff
@@ -224,4 +224,4 @@ ar:
   ZM: زامبيا
   ZW: زمبابوي
 
-  missing: <em>[Translation N/A]</em>
+  missing: ""
```
razorcd: Discussed with Motasem and decided to not show anything when translation is missing.

app/helpers/admin/country_helper.rb
```diff
@@ -3,7 +3,7 @@ module Admin::CountryHelper
   def en_ar_country(country)
     en_country = ISO3166::Country[country]
     ar_country = I18n.with_locale :ar do
-      t country, default: :missing
+        I18n.t country, default: :missing
```
PurityControl: is this indentation correct?

app/helpers/form_helper.rb
```diff
@@ -0,0 +1,10 @@
+module FormHelper
+  def bool_to_int bool
+    bool ? 1 : 0
+  end
+
+  def cancel_button_link record
```
PurityControl: this is sponsor specific and shouldn't be part of helpers as helpers are global.

app/helpers/form_helper.rb
```diff
@@ -0,0 +1,10 @@
+module FormHelper
+  def bool_to_int bool
```
theB264: Do you think FormHelper is the place for this?  It might be.  It could also be `class Object; def bool_to_int; self ? 1 : 0`  See `https://github.com/theB264/osra/blob/4914bf9d25f546753b7d191ed139895cb2509366/spec/ruby_extensions/integer_spec.rb` and `https://github.com/theB264/osra/blob/4914bf9d25f546753b7d191ed139895cb2509366/config/initializers/integer_extension.rb`  And that could allow you to call `f.object.request_fulfilled.bool_to_int.to_s`

app/views/hq/sponsors/_form.html.haml
```diff
@@ -0,0 +1,105 @@
...
+    .form-group
+      = f.label :branch
+      = f.collection_select :branch_id, @branches, :id, :name,
+            {include_blank: true}, class: "form-control"
+  - else
+    .form-group
+      = f.label :request_fulfilled
+      = f.check_box :request_fulfilled, 
+          {disabled: true, checked: f.object.request_fulfilled, class: "checkbox"},
+          "#{bool_to_int f.object.request_fulfilled}"
```
PurityControl: all you need for this is
```
= f.check_box :request_fulfilled, 
     {disabled: true, class: "checkbox"}
```
the `checked:` property is not needed - form_for takes car of that
the value of 1 or 0 is not needed as that is also take care of

app/helpers/form_helper.rb
PurityControl: this method is no longer needed (see my other comment)
I would be opposed to monkey patching core ruby classes when there are other reasonable ways to do this.

PurityControl:
@razorcd Is there a way to highlight trailing white space in your editor? Some of these files have trailing whitespace at the end of some lines.

app/helpers/form_helper.rb
razorcd: Initially I was thinking of adding a `.to_i` method to the `TrueClass` and `FalseClass` but decided to not pollute the ruby classes.

app/views/hq/sponsors/_form.html.haml
razorcd: :+1:

theB264:
Lots of editors will convert all spaces to tabs, ensure trailing line-feed, and remove superfluous trailing white space.  I would look into turning those on

razorcd:
@PurityControl : I set the `trim_trailing_white_space_on_save` in sublime.

app/views/hq/sponsors/edit.html.haml
```diff
@@ -0,0 +1,2 @@
+%h1 Edit Partner #{@sponsor.name}, OSRA number #{@sponsor.osra_num}
+= render partial: 'form', locals: { sponsor: @sponsor, statuses: @statuses, sponsor_types: @sponsor_types, organizations: @organizations, branches: @branches, cities: @cities }
```
razorcd: `_form.html.haml` is independent now. 
And a facade would be very helpful here.

PurityControl: Do you want to tackle that as part of this pr?

razorcd: If you mean creating the facade then no. Definitely different pr.

razorcd:
PR Updated

PR #305, opened 2015-01-23
osra373 make `date-picker` NOT to load all languages

razorcd:
removed loading all language '.js' files except arabic

https://osraav.atlassian.net/browse/OSRA-373


PurityControl:
:+1:  for this

PR #306, opened 2015-01-26
[WIP] The Rails 4 Way - Ch.1-6

NikitaAvvakumov:
I'm currently reading _The Rails 4 Way_ by Obie Fernandez et al., and periodically finding things that could be immediately applied to the OSRA code base. To make note of some smaller tweaks, I've been making corresponding changes on a feature branch you see here. My intent is to share these small modifications and improvements with the team, but this PR is not meant to be mergeable as is. If any of the changes are deemed to be of value, they'll be extracted into their own small PRs.

- generate bundler binstubs. The utility of this may be small, and inclusion of the binstubs in source control is debatable. Skip over all `bin/...` files, just note that they are there.
- add `logger.warn` in partners controller. Recommended as good practice for guard clauses that, ideally, should never trigger - i.e. either the code is broken or the user did something weird or malicious. `WARN` in logs can be easily found and alerts for its appearance there can be set up
- make a few tweaks to `config/routes.rb` that are largely inconsequential but interesting to note
- create a `redirect_to_back_or_default` controller method that can be used in rescue clauses
- add `and return` after each call to `render` or `redirect` in partners controller. Since calling these methods does not automatically terminate the controller action, this is suggested as good practice to avoid accidentally calling them more than once
- make db seeing idempotent. Two different approaches - in the main seeds file, `Province.create(attrs)` (which really ought to be `create!`) replaced with `Province.where(attrs).first_or_create!` - i.e. create only if does not exist. In development seeds, existing records are deleted first and then reseeded. Dev seeds also tweaked to ensure that they don't create additional associated objects.


razorcd:
Great stuff. Thanks for sharing ! :+1:

PR #308, opened 2015-02-01
allow array to paginate as well as an AR instance

theB264:
Extend Array in spec; exception at line 55 supressed

### DIFF
--- a/spec/controllers/hq/partners_controller_spec.rb
+++ b/spec/controllers/hq/partners_controller_spec.rb
@@ -1,4 +1,5 @@
 require 'rails_helper'
+require 'will_paginate/array'
 
 RSpec.describe Hq::PartnersController, type: :controller do
   before :each do


spec/controllers/hq/partners_controller_spec.rb
```diff
@@ -1,4 +1,5 @@
 require 'rails_helper'
+require 'will_paginate/array'
```
PurityControl: why are we including this in the partner controller spec, I can't see where we are currently using it.

theB264:
line 55 tosses an exception if you don't run the whole test suite:
`.and_return([].paginate)`

spec/controllers/hq/partners_controller_spec.rb
PurityControl: What are your thoughts - which is better and why? 
Should we not be returning an array of partners because we are stubbing out the paginate method. Calling paginate on a return value of paginate, to me, seems a little bit misleading.

```
specify 'pagination' do
  partner = build_stubbed :partner
  expect(Partner).to receive_message_chain(:all, :paginate).with(page:
  "2").and_return([partner])
  get :index, page: "2"
end
```

theB264: I'm not sure; this was outside the scope of my story; this is really a bugfix so the develop branch passes build, if you run a spec other than the entire test suite.  But yes; returning an array or activerecord association directly without calling `.paginate` on it first will pop an exception for `undefined method 'total_pages` for ...`

PurityControl: I don't see that this is outside the scope of the bug fix. The bug fix is to make the tests pass in a way that make the tests tell the truth. Both solutions fix the bug but we want to merge in the one thats right.
Currently we have a failing test when run on its own. One fix gets the test to pass but continues to misrepresent what happens. The other fix actually fixes the root cause of the issue and doesn't require an unnecessary require.

theB264: `[partner]` is not something that could be sent to the view.  Exception `undefined method 'total_pages' for ...`  So I'm not sure if using it in the test makes sense.  It would be just as appropriate to return some random object, as it would to return an array or AR_relation.  The view cannot consume an array.

PurityControl: good point

PR #309, opened 2015-02-01
OSRA360 added nav links

theB264:
https://osraav.atlassian.net/browse/OSRA-360


theB264:
screenshot from 2015-02-01 00 25 02

app/views/layouts/navigation.html.erb
```diff
@@ -0,0 +1,31 @@
+<div class="container-fluid">
+  <div class="row">
+    <div class="col-xs-12 col-sm-1 no-horiz-padding">
+      <ul class="nav nav-pills nav-stacked">
+        <% HqController::NAVIGATION_BUTTONS.each do | _, btn | %>
+          <li role="presentation" <%= (request.path=~ btn[:path_regex]) ? 'class=active' : 'class=dormant' %>>
+            <%= link_to %Q{
+              <span class="glyphicon #{btn[:glyph]}" aria-hidden="true"></span>
+              <p class="text-center">#{btn[:text]}</p>
+              }.html_safe, eval(btn[:href]) %>
```
PurityControl: :scream: 
I have never seen html_safe and eval used on the same line together before

PurityControl: I think we need to get to the bottom of why the controller throws a method not found error for the routes rather than trying to work around it.

app/controllers/hq_controller.rb
```diff
@@ -1,4 +1,15 @@
 class HqController < ApplicationController
+  NAVIGATION_BUTTONS= {
```
PurityControl: Navigation order matters so should this not be an array of hashes rather than a hash of hashes?

app/views/layouts/navigation.html.erb
```diff
@@ -0,0 +1,31 @@
+<div class="container-fluid">
+  <div class="row">
+    <div class="col-xs-12 col-sm-1 no-horiz-padding">
+      <ul class="nav nav-pills nav-stacked">
+        <% HqController::NAVIGATION_BUTTONS.each do | _, btn | %>
+          <li role="presentation" <%= (request.path=~ btn[:path_regex]) ? 'class=active' : 'class=dormant' %>>
```
PurityControl: I don't like the idea of a path regex for active classes when we have fails routing functionality we could tap into. If we do this our routing then has to be maintained in the routes file and in the `NAVIGATION_BUTTONS` hash

spec/controllers/hq/hq_controller_spec.rb
```diff
@@ -30,4 +30,11 @@ def foobar_action
       get :foobar_action
     end
   end
+
+  describe 'instantiates' do
+    specify 'navigation buttons' do
+      expect(HqController::NAVIGATION_BUTTONS).to be_present.and be_a_kind_of Hash
```
PurityControl: see my comment re array of hashes

spec/views/layouts/navigation_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      assert_select 'li' do
+        assert_select 'span.glyphicon'
+        assert_select 'span.lorem_ipsum'
+      end
+    end
+
+    specify 'generates link from href code' do
+      stub_const('HqController::NAVIGATION_BUTTONS', {
+        comments: { text: 'foobar', href: '"/pathname/file.extension".reverse', path_regex: /baz/, glyph: 'lorem_ipsum' } })
```
PurityControl: why the call to reverse for the href ?

app/controllers/hq_controller.rb
```diff
@@ -1,4 +1,15 @@
...
+    osra: { text: 'OSRA', href: 'root_path', path_regex: /^\/$/, glyph: 'glyphicon-home' },
+    dashboard: { text: 'Dashboard', href: '"/404.html"', path_regex: /^\/hq(\/)?$/, glyph: 'glyphicon-dashboard' },
+    admin_users: { text: 'Admin Users', href: '"/404.html"', path_regex: /^\/hq\/admin_users/, glyph: 'glyphicon-user' },
+    comments: { text: 'Comments', href: '"/404.html"', path_regex: /^\/hq\/comments/, glyph: 'glyphicon-comment' },
+    orphans: { text: 'Orphans', href: '"/404.html"', path_regex: /^\/hq\/orphans/, glyph: 'glyphicon-user' },
+    partners: { text: 'Partners', href: 'hq_partners_path', path_regex: /^\/hq\/partners/, glyph: 'glyphicon-user' },
+    sponsors: { text: 'Sponsors', href: 'hq_sponsors_path', path_regex: /^\/hq\/sponsors/, glyph: 'glyphicon-user' },
+    users: { text: 'Users', href: 'hq_users_path', path_regex: /^\/hq\/users/, glyph: 'glyphicon-user' }
+  }
```
razorcd: Any reason why you are not using `href: hq_users_path` here and skipping the `eval` in the view?

PurityControl: It throws a method not found error - see my comment above with reference to this.

razorcd: Instead of the `path_regex: /^\/hq\/sponsors/` can't we use something like `controller: 'sponsors'? And then in the view we can check if the `current controller == btn[:controller]`. 
Also, do we care about `hq` here?

razorcd: looks like only `Rails.application.routes.url_helpers.hq_users_path` works here.

PurityControl: @razorcd good call, but I'd like to know why. I would expect these to be available in controllers.

app/views/layouts/navigation.html.erb
theB264: I think this is the same reason `link_to` had to be rigged to work in the parent controller for navigation on my osra262 Rails spike.  I got the hint that these are view helpers and controllers are specifically excluded...

theB264: Yes, it is weird they're on the same line, but the evaluations are being done on constants, and as such, should not contain active content.  But `eval` is scary.

theB264: similar to `validate :foobar, if: 'constant evaluated here in a different scope'`

app/controllers/hq_controller.rb
theB264: Good catch :+1:

spec/views/layouts/navigation_spec.rb
theB264: To test that the constant is being evaluated in-scope as a command rather than, say, using a regex to strip superfluous quotation marks or whatnot

app/controllers/hq_controller.rb
theB264: I think these are called `view helpers`.  Maybe the controller was the wrong place to stash a set of instructions for the views.  Hmm... any ideas where this could be moved where it's testable?

theB264: @razorcd I really like the `controller` suggestion and I will investigate this

razorcd:
I think the path helpers are included only for an instance of the controller or view class?!

@theB264 `app/helpers/navigation.rb` I guess

app/controllers/hq_controller.rb
razorcd: I think the path helpers are included only for an instance of the controller or view class?!

@theB264 `app/helpers/navigation.rb` I guess

razorcd:
one thing bothers me. Why did you move the layout inside `navigation.rb` (header, footer and yield)?
`_navigation.rb` should be a partial and only be included/redered inside `application.rb` layout.

theB264:
@razorcd I put the header and footer inside navigation because I was having trouble getting the view rendering system to work with two `yield`s.  I don't think it's ideal, but they needed to be nested in the tags and I didn't want to split up the opening and closing tags.  This is ripe for change

razorcd:
I looked for a solution to keep the navigation in a partial and made a PR on Brians fork: https://github.com/theB264/osra/pull/2

app/controllers/hq_controller.rb
```diff
@@ -1,4 +1,15 @@
...
+  NAVIGATION_BUTTONS= [
+    { text: 'OSRA', href: 'root_path', path_regex: /^\/$/, glyph: 'glyphicon-home' },
+    { text: 'Dashboard', href: '"/404.html"', path_regex: /^\/hq(\/)?$/, glyph: 'glyphicon-dashboard' },
+    { text: 'Admin Users', href: '"/404.html"', path_regex: /^\/hq\/admin_users/, glyph: 'glyphicon-user' },
+    { text: 'Comments', href: '"/404.html"', path_regex: /^\/hq\/comments/, glyph: 'glyphicon-comment' },
+    { text: 'Orphans', href: '"/404.html"', path_regex: /^\/hq\/orphans/, glyph: 'glyphicon-user' },
+    { text: 'Partners', href: 'hq_partners_path', path_regex: /^\/hq\/partners/, glyph: 'glyphicon-user' },
+    { text: 'Sponsors', href: 'hq_sponsors_path', path_regex: /^\/hq\/sponsors/, glyph: 'glyphicon-user' },
+    { text: 'Users', href: 'hq_users_path', path_regex: /^\/hq\/users/, glyph: 'glyphicon-user' }
+  ]
```
NikitaAvvakumov: What is gained by this approach? The controllers typically link the views to models and/or contain certain amounts of logic that one doesn't want to pollute the view with. Neither is the case here, all that's needed is to render a handful of nav buttons. Why not do everything in the partial?

app/controllers/hq_controller.rb
```diff
@@ -1,4 +1,15 @@
 class HqController < ApplicationController
+  NAVIGATION_BUTTONS= [
+    { text: 'OSRA', href: 'root_path', path_regex: /^\/$/, glyph: 'glyphicon-home' },
+    { text: 'Dashboard', href: '"/404.html"', path_regex: /^\/hq(\/)?$/, glyph: 'glyphicon-dashboard' },
+    { text: 'Admin Users', href: '"/404.html"', path_regex: /^\/hq\/admin_users/, glyph: 'glyphicon-user' },
+    { text: 'Comments', href: '"/404.html"', path_regex: /^\/hq\/comments/, glyph: 'glyphicon-comment' },
```
NikitaAvvakumov: What are 'Dashboard' and 'Comments'?

spec/views/layouts/navigation_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+RSpec.describe 'layouts/navigation.html.erb', type: :view do
+  describe 'renders' do
+    specify 'flashes' do
+      render and expect(view).to render_template 'layouts/_flashes'
+    end
+
+    specify 'footer' do
+      render and expect(view).to render_template 'layouts/_footer'
+    end
+  end
```
NikitaAvvakumov: Why is the above in the navigation partial spec?

spec/controllers/hq/hq_controller_spec.rb
```diff
@@ -30,4 +30,16 @@ def foobar_action
...
+  describe 'instantiates' do
+    specify 'navigation buttons' do
+      expect(HqController::NAVIGATION_BUTTONS).to be_present.and be_a_kind_of Array
+      expect(HqController::NAVIGATION_BUTTONS.first).to be_a_kind_of Hash
+      expect(HqController::NAVIGATION_BUTTONS.first[:text]).to be_present.and be_a_kind_of String
+      expect(HqController::NAVIGATION_BUTTONS.first[:href]).to be_present.and be_a_kind_of String
+      expect(HqController::NAVIGATION_BUTTONS.first[:path_regex]).to be_present.and be_a_kind_of Regexp
+      expect(HqController::NAVIGATION_BUTTONS.first[:glyph]).to be_present.and be_a_kind_of String
+    end
+  end
```
NikitaAvvakumov: Is there much value in this test? Especially considering that it only tests a single button

spec/views/layouts/navigation_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+                  with(HqController::NAVIGATION_BUTTONS.to_a.sample.second[:path_regex]).
+                  and_return(true)
+      render
+    end
+
+    specify 'one higlighted' do
+      assert_select 'div.no-horiz-padding>ul>li.active', {count: 1}
+      assert_select 'div.no-horiz-padding>ul>li.dormant',
+                    {count: (HqController::NAVIGATION_BUTTONS.count - 1)}
+    end
```
NikitaAvvakumov: Is the highlighting functionality provided by Bootstrap? If so, why test it?

theB264: The button highlighting is controlled by bootstrap via class, our app must set this class, which is what this is testing for

theB264:
updated

app/views/hq/partners/index.html.erb
```diff
@@ -1,13 +1,17 @@
-<div class="action_items">
-  <div class="class">
-    <%= link_to('New Partner', new_hq_partner_path,
-                class: 'btn btn-default', role: 'button') %>
+<div class="panel-heading">
+  <div class="action_items">
+    <div class="class">
```
razorcd: should you also remove this `class`here, since you are editing this part anyway?

PR #310, opened 2015-02-01
OSRA342 user index pagination

theB264:
https://osraav.atlassian.net/browse/OSRA-342


theB264:
screenshot from 2015-02-01 02 47 18

spec/views/hq/users/index_spec.rb
```diff
@@ -1,52 +1,62 @@
...
         expect(rendered).to match 'fred.bloggs@example.com'
-        expect(rendered).to match '5'
         expect(rendered).to have_link('Joseph Soap', href: hq_user_path(@users[1]))
         expect(rendered).to match 'joe.soap@example.com'
-        expect(rendered).to match '7'
       end
-    end
 
-  end
+      it 'paginates' do
```
PurityControl: I think this could be changed to 
```
allow(view).to receive(:will_paginate)

render
```
There is no need to artificially return a value from will paginate and check it renders, you are just checking its called.

spec/controllers/hq/partners_controller_spec.rb
```diff
@@ -49,7 +49,7 @@
     end
   end
 
-  describe 'index' do
+  describe '#index' do
     specify 'pagination' do
       expect(Partner).to receive_message_chain(:all, :paginate).with(page: "2").and_return([].paginate)
```
PurityControl: see my other comment in #308  on returning an array rather than `array.paginate` or what you have done for #index in the user controller. I think these are preferable.

app/controllers/hq/users_controller.rb
```diff
@@ -1,6 +1,6 @@
 class Hq::UsersController < HqController
   def index
-    @users = User.all
+    @users = User.all.paginate(page: params[:page])
```
NikitaAvvakumov: Drop the `.all` for consistency with others?

theB264: :+1:

theB264:
updated

spec/controllers/hq/users_controller_spec.rb
```diff
@@ -1,4 +1,5 @@
 require 'rails_helper'
+require 'will_paginate/array'
```
NikitaAvvakumov: Why is this required here but not in the partners_controller_spec?

app/models/user.rb
```diff
@@ -1,4 +1,5 @@
 class User < ActiveRecord::Base
+  self.per_page = 10
```
NikitaAvvakumov: Can this setting be specified elsewhere? I don't like the fact that the AR model now takes on some responsibility for its HTML presentation.

spec/views/hq/users/index_spec.rb
```diff
@@ -1,52 +1,61 @@
...
     context 'no users' do
-      it 'displays just the header row' do
-        @users = []
+      it 'hides the table' do
+        @users = [].paginate
         render
-        expect(rendered).to have_selector('table tr', count: 1)
+
+        expect(rendered).to match /No Users Found/
+        assert_select 'table#users-index', count: 0
```
NikitaAvvakumov: Let's stay with one style of assertions/expectations whenever possible.

spec/views/hq/users/index_spec.rb
```diff
@@ -1,52 +1,61 @@
...
+
         expect(rendered).to have_selector('table tr', count: 2)
         expect(rendered).to have_link('Frederick Bloggs', href: hq_user_path(@users[0]))
         expect(rendered).to match 'fred.bloggs@example.com'
-        expect(rendered).to match '5'
       end
     end
 
     context '>1 user' do
+      before :each do
```
NikitaAvvakumov: Is display of 1 user sufficiently different from display of multiple users to justify two separate tests?

app/models/user.rb
gmunumel: Second that. Maybe put in a file that include public variables that you can take later. Maybe this could help: http://guides.rubyonrails.org/configuring.html#using-initializer-files

theB264:
Don't know why all the PR merge commits.  I did everything, rebase, merge, erase the branch and pull from @razorcd repo, reset back to initial project commit and build up to present, everything.  They still persist.

PR #312, opened 2015-02-02
osra343 controller and view pagination and specs

razorcd:
https://osraav.atlassian.net/browse/OSRA-343

### DIFF
--- a/app/controllers/hq/sponsors_controller.rb
+++ b/app/controllers/hq/sponsors_controller.rb
@@ -1,6 +1,6 @@
 class Hq::SponsorsController < HqController
   def index
-    @sponsors= Sponsor.all
+    @sponsors = Sponsor.paginate(:page => params[:page])
   end
 
   def show
--- a/app/models/sponsor.rb
+++ b/app/models/sponsor.rb
@@ -5,6 +5,7 @@ class Sponsor < ActiveRecord::Base
   PRIORITY_COUNTRIES= %w(SA TR AE GB)
   EXCLUDED_COUNTRYS= %w(IL)
 
+  self.per_page = 10
   attr_accessor :new_city_name
   attr_readonly :branch_id, :organization_id, :sponsor_type_id
 
--- a/app/views/hq/partners/index.html.erb
+++ b/app/views/hq/partners/index.html.erb
@@ -52,6 +52,7 @@
         <%- end %>
       </tbody>
     </table>
+  </div>
   <div class="pagination">
     <%= will_paginate @partners, renderer: BootstrapPagination::Rails %>
   </div>
--- a/app/views/hq/users/show.html.erb
+++ b/app/views/hq/users/show.html.erb
@@ -10,7 +10,7 @@
   <%= pluralize(@user.active_sponsors.count.to_s, 'Active Sponsor') %>
   <% if @user.sponsors.count > 0 %>
     <div class="active_sponsors_index">
-      <%= render partial: 'hq/sponsors/sponsors.html.haml', locals: {sponsors: @user.active_sponsors} %>
+      <%= render partial: 'hq/sponsors/sponsors.html.haml', locals: {sponsors: @user.active_sponsors.paginate(:page => params[:page])} %>
     </div>
   <% end %>
 </div>
@@ -18,7 +18,7 @@
   <%= pluralize(@user.inactive_sponsors.count.to_s, 'Inactive Sponsor') %>
   <% if @user.sponsors.count > 0 %>
     <div class="inactive_sponsors_index">
-      <%= render partial: 'hq/sponsors/sponsors.html.haml', locals: {sponsors: @user.inactive_sponsors} %>
+      <%= render partial: 'hq/sponsors/sponsors.html.haml', locals: {sponsors: @user.inactive_sponsors.paginate(:page => params[:page])} %>
     </div>
   <% end %>
 </div>
--- a/app/views/hq/users/show.html.erb
+++ b/app/views/hq/users/show.html.erb
@@ -10,7 +10,8 @@
   <%= pluralize(@user.active_sponsors.count.to_s, 'Active Sponsor') %>
   <% if @user.sponsors.count > 0 %>
     <div class="active_sponsors_index">
-      <%= render partial: 'hq/sponsors/sponsors.html.haml', locals: {sponsors: @user.active_sponsors.paginate(:page => params[:page])} %>
+      <%= render partial: 'hq/sponsors/sponsors.html.haml',
+            locals: {sponsors: @user.active_sponsors.paginate(:page => params[:page])} %>
     </div>
   <% end %>
 </div>
@@ -18,7 +19,8 @@
   <%= pluralize(@user.inactive_sponsors.count.to_s, 'Inactive Sponsor') %>
   <% if @user.sponsors.count > 0 %>
     <div class="inactive_sponsors_index">
-      <%= render partial: 'hq/sponsors/sponsors.html.haml', locals: {sponsors: @user.inactive_sponsors.paginate(:page => params[:page])} %>
+      <%= render partial: 'hq/sponsors/sponsors.html.haml',
+            locals: {sponsors: @user.inactive_sponsors.paginate(:page => params[:page])} %>
     </div>
   <% end %>
 </div>
--- a/spec/controllers/hq/sponsors_controller_spec.rb
+++ b/spec/controllers/hq/sponsors_controller_spec.rb
@@ -33,21 +33,19 @@ RSpec.describe Hq::SponsorsController, type: :controller do
   end
 
   context '#create' do
-    before :each do
-      @sponsor_new = Sponsor.new
-      @sponsor_build = FactoryGirl.build :sponsor
-    end
-
     specify 'successful' do
-      expect(Sponsor).to receive(:new).and_return(@sponsor_new)
-      expect {post :create, sponsor: @sponsor_build.attributes }.to change{Sponsor.count}.by 1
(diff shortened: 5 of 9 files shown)


razorcd:
**http://localhost:3000/hq/sponsors?page=4**

sponsor index_pagination

__________________________________________________________________________________
**http://localhost:3000/hq/users/1**

user index_sponsors_pagination

spec/controllers/hq/sponsors_controller_spec.rb
```diff
@@ -1,16 +1,19 @@
...
+require 'will_paginate/array'
 
 RSpec.describe Hq::SponsorsController, type: :controller do
   before :each do
     sign_in instance_double(AdminUser)
     @sponsor = FactoryGirl.build_stubbed :sponsor
+    @sponsors = (1..5).map {FactoryGirl.build_stubbed :sponsor}
     @sponsor_build = FactoryGirl.build :sponsor
     @sponsor_new = Sponsor.new
   end
```
razorcd: Permission to change all these instance variables into `let` otherwise all will be loaded even when not necessary.

app/controllers/hq/sponsors_controller.rb
```diff
@@ -1,6 +1,6 @@
 class Hq::SponsorsController < HqController
   def index
-    @sponsors= Sponsor.all
+    @sponsors = Sponsor.paginate(:page => params[:page])
```
razorcd: Notice: I didn't use `Sponsor.all.paginate(:page => params[:page])` like in the Partner#index controller. They both result in the same SQL query.

spec/controllers/hq/sponsors_controller_spec.rb
theB264: I've never been a fan of using instance variables here from the beginning

app/controllers/hq/sponsors_controller.rb
theB264: I think they would, unless you set up a default scope on your model.  Then `Class.paginate` might be a different query.  I would prefer to see us use one or the other, in other words, changing Partner to not use `.all` in this PR or changing Sponsor to use `.all`

app/views/hq/partners/index.html.erb
```diff
@@ -50,6 +50,7 @@
         <%- end %>
       </tbody>
     </table>
+  </div>
```
theB264: LoL; I fixed this too in my WIP

app/views/hq/users/show.html.erb
```diff
@@ -18,7 +18,7 @@
   <%= pluralize(@user.inactive_sponsors.count.to_s, 'Inactive Sponsor') %>
   <% if @user.sponsors.count > 0 %>
     <div class="inactive_sponsors_index">
-      <%= render partial: 'hq/sponsors/sponsors.html.haml', locals: {sponsors: @user.inactive_sponsors} %>
+      <%= render partial: 'hq/sponsors/sponsors.html.haml', locals: {sponsors: @user.inactive_sponsors.paginate(:page => params[:page])} %>
```
theB264: This should probably be discussed; do we want both copies of the view partial paginating to the same place?  I would enter a task to uncouple these under osra327, whether that can be done directly or if it means separating the partial into two partials, one for Sponsor and one for User.  Another option would be to not paginate these as each user probably wont have an absurd amount of Sponsors

app/views/hq/users/show.html.erb
```diff
@@ -18,7 +19,8 @@
   <%= pluralize(@user.inactive_sponsors.count.to_s, 'Inactive Sponsor') %>
   <% if @user.sponsors.count > 0 %>
     <div class="inactive_sponsors_index">
-      <%= render partial: 'hq/sponsors/sponsors.html.haml', locals: {sponsors: @user.inactive_sponsors} %>
+      <%= render partial: 'hq/sponsors/sponsors.html.haml',
+            locals: {sponsors: @user.inactive_sponsors.paginate(:page => params[:page])} %>
```
razorcd: User#show also uses the `_sponsors` partial here. 
To overwrite the default number of sponsors per page we ca use:
`@user.inactive_sponsors.paginate(:page => params[:page], :per_page => 15)}`

theB264: If you use `params[:page]` it will lock both of them together; is there a solution involving something like `params[:inactive][:page]` but then that needs to be in the links in `_sponsors` ...

spec/controllers/hq/sponsors_controller_spec.rb
```diff
@@ -1,38 +1,47 @@
...
+    before :each do
+      @sponsor_new = Sponsor.new
+      @sponsor_build = FactoryGirl.build :sponsor
+    end
+
     specify 'successful' do
       expect(Sponsor).to receive(:new).and_return(@sponsor_new)
       expect {post :create, sponsor: @sponsor_build.attributes }.to change{Sponsor.count}.by 1
-      expect(response).to redirect_to hq_sponsor_url(Sponsor.find_by_name(@sponsor_build.name))
+      expect(response).to redirect_to hq_sponsor_url( Sponsor.find_by_name(@sponsor_build.name) )
```
NikitaAvvakumov: I realize this is not a part of this PR, but the test on L43 does not seem to belong in a controller spec. It exercises not only the controller but the model and the db as well - something that should be reserved for integration-level testing. What are people's thoughts?

theB264: I think making a task to move that expectation into an RSpec integration test or possibly a Cucumber integration test is a good idea.  But this PR probably not

spec/controllers/hq/sponsors_controller_spec.rb
```diff
@@ -30,38 +34,41 @@
...
-      expect {post :create, sponsor: @sponsor_build.attributes }.to change{Sponsor.count}.by 1
-      expect(response).to redirect_to hq_sponsor_url(Sponsor.find_by_name(@sponsor_build.name))
+      expect_any_instance_of(Sponsor).to receive(:save) do |_self|
+        _self.id = sponsor.id
+        true
+      end
+
+      post :create, sponsor: sponsor.attributes
+      expect(response).to redirect_to hq_sponsor_url(sponsor)
     end
```
razorcd: and we are back at mocking the save :D

app/models/sponsor.rb
```diff
@@ -5,6 +5,7 @@ class Sponsor < ActiveRecord::Base
   PRIORITY_COUNTRIES= %w(SA TR AE GB)
   EXCLUDED_COUNTRYS= %w(IL)
 
+  self.per_page = 10
```
NikitaAvvakumov: As in my comment on #310 - can we find a different place to store these settings?

app/views/hq/users/show.html.erb
```diff
@@ -10,7 +10,8 @@
   <%= pluralize(@user.active_sponsors.count.to_s, 'Active Sponsor') %>
   <% if @user.sponsors.count > 0 %>
```
NikitaAvvakumov: Should this be `@user.active_sponsors.count`? And `.inactive_sponsors` on L20?

spec/controllers/hq/sponsors_controller_spec.rb
```diff
@@ -30,38 +34,41 @@
 
   context '#create' do
     specify 'successful' do
-      expect(Sponsor).to receive(:new).and_return(@sponsor_new)
-      expect {post :create, sponsor: @sponsor_build.attributes }.to change{Sponsor.count}.by 1
-      expect(response).to redirect_to hq_sponsor_url(Sponsor.find_by_name(@sponsor_build.name))
+      expect_any_instance_of(Sponsor).to receive(:save) do |_self|
```
NikitaAvvakumov: Can we use the approach from #283 and avoid this?

app/views/hq/users/show.html.erb
theB264: I had specifically changed this from that; currently either both partials render or neither, but not just one.  This is a UI decision for the group

theB264: The tests that were written with this will allow either one, this is a style thing

NikitaAvvakumov: I was just curious. It's nothing to hold up the PR for.

app/models/sponsor.rb
razorcd: lets put this in a new task so we can change all models

spec/controllers/hq/sponsors_controller_spec.rb
```diff
@@ -30,38 +34,36 @@
 
   context '#create' do
     specify 'successful' do
-      expect(Sponsor).to receive(:new).and_return(@sponsor_new)
-      expect {post :create, sponsor: @sponsor_build.attributes }.to change{Sponsor.count}.by 1
-      expect(response).to redirect_to hq_sponsor_url(Sponsor.find_by_name(@sponsor_build.name))
+      post :create, sponsor: build(:sponsor).attributes
+      expect(response).to redirect_to hq_sponsor_url(assigns :sponsor)
     end
```
razorcd: Why I used `build(:sponsor).attributes`.
Because an `agent` is needed for saving the new sponsor.

```ruby
ss = Sponsor.new( FactoryGirl.build(:sponsor).attributes ); ss.save    ->  true

ss = Sponsor.new( FactoryGirl.build_stubbed(:sponsor).attributes ); ss.save    ->  false
ss.errors   ->    @messages={:agent=>["can't be blank"]}>

ss = Sponsor.new( FactoryGirl.attributes_for(:sponsor) ); ss.save      -> false
ss.errors   ->    @messages={:agent=>["can't be blank"]}>
```

PR #314, opened 2015-02-04
Consolidate sponsorship-related actions

NikitaAvvakumov:
The Sponsorship model was taking too much responsibility for managing its orphan's sponsorship status. It was also doing this in a redundant fashion. This small update removes the redundancy and hands all responsibility for determining sponsorship status to the orphan.

As an added bonus, since orphan status change upon sponsorship inactivation is now invoked within sponsorship's `after_save` callback rather than from the `inactivate` method, these two events are automatically wrapped in a db transaction and thus become atomic.

 - consolidate calls to sponsor & orphan update methods within Sponsorship `after_save` & `after_destroy`
 - implement `Orphan#sponsorship_changed!` which is called from Sponsorship `after_save` & `after_destroy`
 - rename `Orphan#reactivate` to `#resolve_sponsorship_status` to reflect its use both upon orphan reactivation and sponsorship changes
 - remove methods made redundant by the above changes
 - adjust specs accordingly

### DIFF
--- a/app/models/orphan.rb
+++ b/app/models/orphan.rb
@@ -124,8 +124,8 @@ class Orphan < ActiveRecord::Base
     current_sponsorship.sponsor if currently_sponsored?
   end
 
-  def resolve_sponsorship_status
-    reactivate and save!
+  def sponsorship_changed!
+    resolve_sponsorship_status and save!
   end
 
 private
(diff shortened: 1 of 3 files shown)


lib/tasks/clear_orphans.rake
```diff
@@ -5,7 +5,7 @@ namespace :db do
     OrphanList.destroy_all
     PendingOrphan.destroy_all
     PendingOrphanList.destroy_all
-    Sponsorship.destroy_all
+    Sponsorship.delete_all
```
PurityControl: do we need to rebase to keep the comment ?

theB264:
:shipit:

PR #318, opened 2015-02-05
OSRA359: Added header to the page

gmunumel:
Added the header to the web page. 

screenshot from 2015-02-05 20 36 52

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -1,0 +1,15 @@
+<nav class="osra-nav-header" role="navigation">
+  <div class="osra-header-inner">
+      <div class="osra-header-primary">
+        <p>OSRA Support System</p>
+      </div>
+      <div class="osra-header-secondary">
+        <ul class="osra-header __secondary">
+          <li>example@example</li>
+          <li>logout</li>
+        </ul>
+      </div>
+  </div>
+</nav>
+<div class="osra-header-clear">
+
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -1,17 +1,17 @@
-<nav class="osra-nav-header" role="navigation">
-  <div class="osra-header-inner">
-      <div class="osra-header-primary">
-        <p>OSRA Support System</p>
-      </div>
-      <div class="osra-header-secondary">
-        <ul class="osra-header __secondary">
-          <% if admin_user_signed_in?  %>
-            <li> <%= link_to current_admin_user.email, admin_admin_user_path(current_admin_user.id) %></li>
-            <li> <%= link_to 'logout', :destroy_admin_user_session %> </li>
-          <% end %>
-        </ul>
-      </div>
-  </div>
-</nav>
-<div class="osra-header-clear">
+<nav class="navbar navbar-default osra-nav-header">
+  <div class="container-fluid">
+    <!-- Brand and toggle get grouped for better mobile display -->
+    <div class="navbar-header">
+      <a class="navbar-brand" href="#">OSRA Support System</a>
+    </div>
 
+    <!-- Collect the nav links, forms, and other content for toggling -->
+    <div class="collapse navbar-collapse" id="bs-example-navbar-collapse-1">
+      <ul class="nav navbar-nav navbar-right">
+        <% if admin_user_signed_in?  %>
+          <li> <%= link_to current_admin_user.email, admin_admin_user_path(current_admin_user.id) %></li>
+          <li> <%= link_to 'logout', :destroy_admin_user_session %> </li>
+        <% end %>
+      </ul>
+  </div><!-- /.container-fluid -->
+</nav>
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -5,8 +5,10 @@
       </div>
       <div class="osra-header-secondary">
         <ul class="osra-header __secondary">
-          <li>example@example</li>
-          <li>logout</li>
+          <% if admin_user_signed_in?  %>
+            <li> <%= link_to current_admin_user.email, admin_admin_user_path(current_admin_user.id) %></li>
+            <li> <%= link_to 'logout', :destroy_admin_user_session %> </li>
+          <% end %>
         </ul>
       </div>
   </div>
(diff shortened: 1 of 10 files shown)


app/views/layouts/_header.html.erb
```diff
@@ -0,0 +1,17 @@
+<nav class="osra-nav-header" role="navigation">
+  <div class="osra-header-inner">
+      <div class="osra-header-primary">
+        <p>OSRA Support System</p>
+      </div>
+      <div class="osra-header-secondary">
+        <ul class="osra-header __secondary">
+          <% if admin_user_signed_in?  %>
```
theB264: We should probably have a test to check this is working

gmunumel: I can do that, however, the difficult part is to test when the user is already logged in, that should be the next part.

app/views/layouts/_header.html.erb
```diff
@@ -0,0 +1,17 @@
+<nav class="osra-nav-header" role="navigation">
```
NikitaAvvakumov: Can we use Bootstrap's `navbar` classes here instead and avoid having to create and fine-tune our own?

spec/support/controller_helpers.rb
```diff
@@ -13,5 +13,6 @@ def sign_in(user= instance_double(AdminUser))
 
 RSpec.configure do |config|
   config.include Devise::TestHelpers, :type => :controller
+  config.include Devise::TestHelpers, :type => :view
```
NikitaAvvakumov: Should we put this in a "view_helpers.rb"? Or rename this file?

spec/factories/admin_user.rb
```diff
@@ -0,0 +1,7 @@
+FactoryGirl.define do
+  factory :admin_user do
+    email "admin@example.com"
+    password "password"
+    password_confirmation {|u| u.password}
+  end
+end
```
NikitaAvvakumov: Can you create a simple spec for the model & test the factory in it as well?

spec/views/layouts/_header_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+    @admin = FactoryGirl.create :admin_user
+    sign_in @admin
+  end
+
+  it 'should show header information' do
+    render
+    expect(rendered).to have_link(@admin.email, href: admin_admin_user_path(@admin.id))
+    expect(rendered).to have_link('logout', href: destroy_admin_user_session_path)
+  end
+end
```
razorcd: Nice work figuring this out :+1:

app/views/layouts/application.html.erb
```diff
@@ -9,6 +9,7 @@
   </head>
   <body>
     <div class="container-fluid">
+      <%= render 'layouts/header' %>
```
theB264: If you rebase your PR, a new file will show up, `spec/views/layouts/application_spec.rb`
We should probably include this render call in the spec

gmunumel: Hello @theB264 , I tried to rebase my PR, but I don't see the application_spec.rb file. I follow the Nik's guide of git but it didn't work. Can you help me with this?

gmunumel:
This is how it looks now the header after added the `bootsrap's navbar`
screenshot from 2015-02-13 11 27 06
.

spec/models/admin_user_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  it 'should have a valid factory' do
+    expect(build_stubbed :admin_user).to be_valid
+  end
+
+  it { is_expected.to validate_presence_of :email }
+  it { is_expected.to validate_presence_of :password }
+  it { is_expected.not_to validate_presence_of :password_confirmation }
+  it { is_expected.to validate_uniqueness_of :email }
+  it { is_expected.not_to validate_uniqueness_of :password }
+  it { is_expected.not_to validate_uniqueness_of :password_confirmation }
```
NikitaAvvakumov: Can dispense with the negative tests on L11, 13 & 14.

NikitaAvvakumov: In fact, if I'm reading Devise docs correctly, we can remove all tests except the one for the factory. `AdminUser` includes Devise's `validatable` module which validates email presence, uniqueness & format, as well as password presence, length (8-72) & confirmation. A single test to ensure that the module is included is sufficient; more than that and we are testing the gem.

spec/factories/admin_users.rb
```diff
@@ -0,0 +1,7 @@
+FactoryGirl.define do
+  factory :admin_user do
+    email { Faker::Internet.email }
```
NikitaAvvakumov: We've had problems before with Faker generating non-unique values in a single run of the test suite. To avoid this, we've been appending a sequential number to any Faker-generated attributes that must be unique. See the partners factory, e.g.

spec/models/admin_user_spec.rb
gmunumel: Yes, I thought of that. But I do it anyways thinking that may be in the future I will have to get rid of them.

spec/factories/admin_users.rb
```diff
@@ -0,0 +1,7 @@
+FactoryGirl.define do
+  factory :admin_user do
+    email { [Faker::Internet.email].sample }
```
NikitaAvvakumov: Sorry, I meant the other one - `sequence :email { |n| "#{n}#{Faker::Internet.email}" }`. Another option is to get rid of Faker completely here: `sequence :email { |n| "admin#{n}@example.com" }`.

The `sequence` is the key thing here: we've found that e.g. `Faker::Internet.email` can sometimes return the same value on consecutive calls. So, if a test file tries to create two AdminUsers and they both get the same Faker email, the second record will be invalid since its email won't be unique. Appending a sequential number ensures that even if Faker returns e.g. "abc@this.com" twice, the final email won't be the same: "1abc@this.com", "2abc@this.com" etc.

gmunumel: I tried putting in this way `sequence :email { |n| "#{Faker::Internet.email} #{n}" }` but give an error,  related to what you are talking about. I will try as you commented.

NikitaAvvakumov: The problem with your approach is that email format is validated by Devise. `"#{Faker::Internet.email} #{n}"` would generate e.g. "admin@example.com 1", which would fail the validation. The sequence number must be appended to the front of the address or internally, but without interfering with the email's domain format.

gmunumel: Exactly :+1:

PR #322, opened 2015-02-13
pt.I: InactivateSponsorship service object

NikitaAvvakumov:
### Part I of IV - replace model callbacks handling sponsorship events with service objects.

Here, a `InactivateSponsorship` PORO class is created to handle sponsorship inactivation and attendant changes to the associated sponsor & orphan records. The sponsorships controller (AA) is modified to use this service object instead of calling sponsorship, sponsor & orphan instance methods.

Most of this code will change in parts 2-4, so focus on the approach rather than the implementation.

### DIFF
--- a/app/services/inactivate_sponsorship.rb
+++ b/app/services/inactivate_sponsorship.rb
@@ -1,0 +1,43 @@
+class InactivateSponsorship
+
+  def initialize(sponsor:, sponsorship:, end_date:)
+    @sponsorship = sponsorship
+    @end_date = end_date
+    @sponsor = sponsor
+  end
+
+  def call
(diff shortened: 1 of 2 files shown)


app/services/inactivate_sponsorship.rb
```diff
@@ -0,0 +1,43 @@
+class InactivateSponsorship
+
+  def initialize(sponsor:, sponsorship:, end_date:)
```
theB264: the keyword arguments are a nice touch

theB264:
:+1:

NikitaAvvakumov:
Part IV merged

PR #323, opened 2015-02-13
pt. II CreateSponsorship service object

NikitaAvvakumov:
### Part II of IV - replace model callbacks handling sponsorship events with service objects.

Here, a `CreateSponsorship` service object is created. Given the overlap between it and `InactivateSponsorship` from pt. I, common parts are extracted into their own small service objects: `UpdateOrphanSponsorshipStatus` & `UpdateSponsorSponsorshipData`. All changes to sponsorship, sponsor & orphan are wrapped in an AR transaction to ensure atomic execution.

The common pattern for all service objects is to have a `#initialize` method that sets requisite instance variables, and a `#call` method that executes the service object's key function. The usage syntax is, therefore, `ServiceObject.new(params).call`.

Again, this code will change in pts. 3 & 4, so pay most attention to the approach rather than implementation details.

### DIFF
--- a/app/admin/sponsorship.rb
+++ b/app/admin/sponsorship.rb
@@ -4,14 +4,18 @@ ActiveAdmin.register Sponsorship do
   belongs_to :sponsor
 
   member_action :inactivate, method: :put do
-    sponsorship = Sponsorship.find(params[:id])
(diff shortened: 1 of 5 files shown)


NikitaAvvakumov:
Part IV merged

PR #324, opened 2015-02-13
pt. III DestroySponsorship service object

NikitaAvvakumov:
### Part III of IV - replace model callbacks handling sponsorship events with service objects.

Here, a `DestroySponsorship` service object is created following the patterns established in pts. I & II. Since upon sponsorship destruction an orphan's sponsorship status must be resolved based on its remaining sponsorship associations, a `ResolveOrphanSponsorshipStatus` service object is also created. This functionality is also used whenever an orphan is reactivated (status change from 'On Hold' or 'Inactive' to 'Active') - see pt. IV.

### DIFF
--- a/app/admin/sponsorship.rb
+++ b/app/admin/sponsorship.rb
@@ -36,5 +36,19 @@ ActiveAdmin.register Sponsorship do
     end
   end
 
+  def destroy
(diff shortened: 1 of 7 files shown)


NikitaAvvakumov:
Part IV merged

PR #325, opened 2015-02-13
pt. IV Sponsorship Service Objects

NikitaAvvakumov:
### Part IV of IV - replace model callbacks handling sponsorship events with service objects.

Here, callbacks and related code are removed from the models as they have now been fully replaced with service objects. Specs & features are updated.

I was initially concerned by the fact that a sponsorship can no longer be properly created via a factory or `Sponsorship.create` - without callbacks in the model, the attendant sponsor & orphan records won't be updated to reflect their association with the sponsorship. However, I've come to see this as much less of a problem over time (if it is a problem at all). The factory and `Sponsorship.create` still produce valid sponsorship objects, which means that the Sponsorship model fully encapsulates its own functionality. Updating sponsor & orphan is, to me, clearly beyond the scope of Sponsorship's responsibilities, and so relying on service objects to effect all the changes that need to occur every time a sponsorship changes feels right to me.

### DIFF
--- a/app/admin/sponsorship.rb
+++ b/app/admin/sponsorship.rb
@@ -1,33 +1,9 @@
 ActiveAdmin.register Sponsorship do
   menu if: proc { false }
-  actions :create
+  actions :create, :destroy
   belongs_to :sponsor
 
   member_action :inactivate, method: :put do
-    sponsor = Sponsor.find(params[:sponsor_id])
-    sponsorship = sponsor.sponsorships.find(params[:id])
-    begin
-      InactivateSponsorship.new(sponsorship: sponsorship,
-                                end_date: params[:end_date]).call
-      flash[:success] = 'Sponsorship link was successfully terminated.'
-    rescue => error
-      flash[:warning] = error.message
-    ensure
-      redirect_to admin_sponsor_path(sponsor)
-    end
-  end
-
-  member_action :destroy, method: :delete do
-    sponsor = Sponsor.find(params[:sponsor_id])
-    sponsorship = sponsor.sponsorships.find(params[:id])
-    begin
-      DestroySponsorship.new(sponsorship).call
-      flash[:success] = 'Sponsorship record was successfully destroyed.'
-    rescue => error
-      flash[:warning] = error.message
-    ensure
-      redirect_to admin_sponsor_path(sponsor)
-    end
   end
 
   controller do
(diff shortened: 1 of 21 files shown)


app/services/create_sponsorship.rb
```diff
@@ -0,0 +1,34 @@
+class CreateSponsorship
+
+  attr_reader :sponsorship, :sponsor, :orphan
```
razorcd: do we need the `attr_reader` on services? If we use them in code (eg in controllers) we might just complicate things.
Or they are here only for debugging?
Shouldn't services be used only to perform a set of related tasks and nothing else?

app/admin/sponsorship.rb
```diff
@@ -1,30 +1,47 @@
...
         flash[:success] = 'Sponsorship link was successfully created.'
-        redirect_to admin_sponsor_path(@sponsor.id)
-      else
-        flash[:warning]= sponsorship.errors.full_messages || 'Sponsorship not created'
-        redirect_to new_sponsorship_path(@sponsor.id, scope: 'eligible_for_sponsorship')
+        redirect_to admin_sponsor_path(sponsor)
+      rescue => error
+        flash[:warning] = error.message
+        redirect_to new_sponsorship_path(sponsor, scope: 'eligible_for_sponsorship')
       end
```
razorcd: personally, I would prefer the `.call` to return true or false instead of using raise-rescue every time.

razorcd:
The code is nice and clean. I like that. It adds all related tasks in one place, this is a lit easier to read/maintain.
I am just wondering, instead of using classes to create the services, isn't it better to just create modules and include them in the Model?

app/services/create_sponsorship.rb
NikitaAvvakumov: The point of using `attr_reader` here is not to provide public access to the instance variables but to encapsulate access to them within the class itself. These accessors can - and perhaps should - be made private.

app/admin/sponsorship.rb
NikitaAvvakumov: The raise-rescue approach here was taken because of AR transactions that wrap the code in `#call`. For a transaction to be rolled back in case something goes wrong, the methods within it have to throw errors rather than return `false`. It is possible to rescue within `#call` itself and return `false` to the controller, but then propagating the error messages to the flash would be more convoluted.

Constant reliance on `rescue` is not my preferred option either, perhaps we can find something better.

NikitaAvvakumov:
I feel that moving code into modules and then mixing those into single classes accomplished little - you just split a model into multiple files, but ultimately everything remains the same. Here, Sponsorship would still be responsible for updating Sponsor & Orphan objects, and the latter two would still accumulate logic that goes beyond an AR model's core responsibilities. Plus, you'd then have to hunt for method definitions in the modules and not only the models themselves. With service objects, bits of functionality are better encapsulated and SOLID principles are followed more closely.

Since mixing in modules is Ruby's solution to multiple inheritance, the module approach would also run counter to the "Prefer Composition over Inheritance" mantra. Sponsorship would then effectively be a child of e.g. CreateSponsorship module as it would include all of the latter's methods. By contrast, a CreateSponsorship service object is _composed_ of sponsorship, sponsor and orphan objects which get assigned to its instance variables. It can both define its own methods and call those defined on sponsorship et al., which significantly increases flexibility.

razorcd:
yes, I know what you mean. A module would just group the methods in a new file. But still we could just have a public method that we just call from the controller and the rest would be private methods. It is not clean code but it would be easier to use in the controller, I think.

To be honest what bothers me is that for every service, we have to create a new object (service) that just does a few tasks and then is destroyed by GC.  Also the controller got bigger and we have to remember a different syntax for working with the model.

app/admin/sponsorship.rb
razorcd: this also means we have to repeat the `rescue` over and over again. 
For validation rails is using the `sponsorship.errors` but that is returning a hash for each invalid db column, we can't use that to return the error message, can we? Or can we do something similar?

app/services/create_sponsorship.rb
razorcd: you can use `@sponsorship.save!` instead of `sponsorship.save!`. Not sure which is more preferred.

razorcd:
After just reading about the Command Design Pattern, looks like this is the right way to run a bunch of tasks... use a Class for all the tasks and then create a new object and execute. lesson learned! :)

spec/controllers/admin/sponsorships_controller_spec.rb
```diff
@@ -3,30 +3,74 @@
...
+    end
+
+    context 'when successful' do
 
-      it 'calls #inactivate on sponsorship' do
-        expect(sponsorship).to have_received :inactivate
+      before(:each) do
+        expect(DestroySponsorship).to receive_message_chain(:new, :call).
+          and_return true
+        delete :destroy, { id: 1, sponsor_id: 1 }
```
theB264: I wonder if we still need the sponsor_id in the route when a sponsorship_id is enough information to resolve that.  Out of the scope of this PR though

NikitaAvvakumov: That occurred to me too.

theB264: smells like a jira task

gmunumel:
I really like the code.  It seems it took a lot work. 

:+1:

spec/controllers/admin/sponsorships_controller_spec.rb
NikitaAvvakumov: The only param expected by the `DestroySponsorship` service is sponsorship. We can leave the route as is in AA and implement it more efficiently in Rails without having to refactor the service.

NikitaAvvakumov:
@razorcd Following our discussion, I moved error rescue out of the controller and into the services, which I think improves both. Services now return hashes in a format mergeable with the flash, e.g. `{ :success => 'Sponsorship created' }`. This ties them to the controller somewhat, but this is probably not a big issue since that error message is perfectly readable outside of the flash as well. Let me know what you think.

NikitaAvvakumov:
@razorcd Thanks for your further feedback! I've made the suggested modifications, and the result indeed looks cleaner and is much more idiomatic Rails than the previous implementation.

I wasn't sure what the `#undo` method would do in these services. The transaction rolls back if any of the wrapped methods fails and raises an error (and I tried to be diligent and use bang methods everywhere for this reason), so there is no need to undo anything if `#call` returns `false`.

app/admin/sponsorship.rb
```diff
@@ -4,27 +4,67 @@
...
+      end
+    end
+
+    def destroy_sponsorship
+      if @sponsorship_destructor.call
+        flash[:success] = 'Sponsorship record was successfully destroyed.'
       else
-        flash[:warning]= sponsorship.errors.full_messages || 'Sponsorship not created'
-        redirect_to new_sponsorship_path(@sponsor.id, scope: 'eligible_for_sponsorship')
+        flash[:warning] = @sponsorship_destructor.error_msg
```
razorcd: we only accept `:success` and `:error` flash messages` in rails. Should we add warning too?

NikitaAvvakumov: I think we can keep `:warning` in AA and transition it to `:error` in HQ

gmunumel: We can add it, but the idea of use it. Otherwise, it will be useless.

PR #327, opened 2015-02-15
Improve Sponsor#currently_sponsored_orphans

NikitaAvvakumov:
Change implementation of `Sponsor#currently_sponsored_orphans` method to
retrieve the orphan records entirely via a db query and to return an AR
relation rather than array


gmunumel:
Like, because you can play more with the AR rather than with the array. I guess is there a motivation for this change.

NikitaAvvakumov:
@gmunumel That was the exact reason

PR #330, opened 2015-02-17
osra376 new regex for email validation

razorcd:
https://osraav.atlassian.net/browse/OSRA-376

Play with the regex here: http://rubular.com/r/PFwowZvaxs

### DIFF
--- a/app/models/sponsor.rb
+++ b/app/models/sponsor.rb
@@ -30,7 +30,9 @@ class Sponsor < ActiveRecord::Base
   validate :date_not_beyond_first_of_next_month
   validate :belongs_to_one_branch_or_organization
   validate :can_be_inactivated, if: :being_inactivated?, on: :update
-  validates_format_of :email, with: Devise.email_regexp, allow_blank: true
+  validates_format_of :email,
+            with: /\A[_a-z0-9](\.?[\-+_%a-z0-9]+)*@([_+\-%a-z0-9]+)(\.[_+\-%a-z0-9]+)*(\.[_a-z0-9]+)+\z/i,
+            allow_blank: true
   validate :type_matches_affiliation, on: :create
   validates :agent, presence: true
 
--- a/app/models/sponsor.rb
+++ b/app/models/sponsor.rb
@@ -31,8 +31,8 @@ class Sponsor < ActiveRecord::Base
   validate :belongs_to_one_branch_or_organization
   validate :can_be_inactivated, if: :being_inactivated?, on: :update
   validates_format_of :email,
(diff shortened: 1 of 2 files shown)


app/models/sponsor.rb
```diff
@@ -30,7 +30,9 @@ class Sponsor < ActiveRecord::Base
   validate :date_not_beyond_first_of_next_month
   validate :belongs_to_one_branch_or_organization
   validate :can_be_inactivated, if: :being_inactivated?, on: :update
-  validates_format_of :email, with: Devise.email_regexp, allow_blank: true
+  validates_format_of :email,
+            with: /\A[_a-z0-9](\.?[\-+_%a-z0-9]+)*@([_+\-%a-z0-9]+)(\.[_+\-%a-z0-9]+)*(\.[_a-z0-9]+)+\z/i,
+            allow_blank: true
   validate :type_matches_affiliation, on: :create
```
razorcd: Devise: `email_regexp` should be used to give user feedback and not to assert the e-mail validity.
See here:
https://github.com/plataformatec/devise/blob/c19f1f27cecd7024df0ff8b4ace400d665dd9576/lib/devise.rb#LC119

razorcd:
I got this from Brian:
http://davidcel.is/blog/2012/09/06/stop-validating-email-addresses-with-regex/

Having a bullet proof regex to validate the email is almost impossible.

We could add a "confirm email" field to the `new` (and `edit`?) and the send the sponsor a confirmation/welcome email. We will need another field `email_confirmed` : boolean

Or we can keep the regex for now, it will cover most of the fake emails.

NikitaAvvakumov:
The situation is less clear-cut than e.g. what is described in the article. Email is not a required attribute for sponsors, from which I conclude that OSRA do not communicate with sponsors exclusively via email. Thus, email verification would not be a natural part of their communication strategy, but something extra - e.g. they might verify a sponsor's email but then never use it to contact him/her if the sponsor indicates a preference for phone conversations. 

Furthermore, there is a big difference between verifying the emails of your site's users who probably use them for core functionalities like login, receiving periodic updates about tweets/pins/likes etc. and doing so in cases where email is collected simply as part of general information about a person who will not be using the site personally. In this latter case, email is no different than phone numbers, addresses or even names, which are not verified.

We can run the idea of verification past the client, but it doesn't feel like the best fit here.

As an aside, based on [this discussion](http://stackoverflow.com/a/150078/2623544), `[\u00C0-\u1FFF\u2C00-\uD7FF\w]` should cover all alphabets, so it can be used in place of simple `\w`.

razorcd:
`[\u00C0-\u1FFF\u2C00-\uD7FF\w]` works too but I used `[[:word:]]` because it is shorter.
The new regex will cover all foreign characters and some accepted characters(see tests).

play with it here:  http://rubular.com/r/JPV8QcFJvg

gmunumel:
I tested with some chinese characters. It works great. It cannot be perfect. But it's a very good starting point.

PR #331, opened 2015-02-18
Gem update

NikitaAvvakumov:
- Update all gems to latest versions, including Rails to 4.2.0.
   - Add web-console for debugging
   - Add missing comments to Gemfile
   - Add pessimistic version locks to all gems
   - Add configuration options to address deprecation warnings
   - Dump new schema.rb
   - Generate binstubs for all executables.
   - Fix 2 RSpec failures
   - Fix 1 Cucumber failure


app/views/admin/pending_orphan_lists/valid_list.html.arb
```diff
@@ -4,7 +4,7 @@ div do
   table_for result do
     column :name
     column :father_name
-    column :date_of_birth
+    column(:date_of_birth) { |orphan| format_date(Date.parse(orphan.date_of_birth)) }
```
NikitaAvvakumov: Instead of conforming to Rails' new default date format, convert orphans' DOBs to the common full-date format as specified by the client.

gmunumel:
How often is going to be the Gem update?

NikitaAvvakumov:
Something like: we upgrade for significant Ruby and Rails releases, and update all gems at/around the same time. Otherwise, we update individual gems if Gemnasium raises an alert or we discover that new functionality or fix is needed by our code base. This is open for discussion.

gmunumel:
Got it! :+1: 

We need to be careful of the backward compatibility

PR #333, opened 2015-02-19
OSRA-120 Create and Add Another Sponsor (HQ)

NikitaAvvakumov:
https://osraav.atlassian.net/browse/OSRA-120

Add 'Create and Add Another' button to new sponsor page _in the HQ namespace_ (AA still to come).
 - add conditionally rendered button to the sponsor form
 - modify `sponsors_controller#save_sponsor` to parse `params[:commit]`
 - add button tests to new & edit view specs
 - create RSpec feature to test new record creation & redirect to new sponsor view - discuss!
 - add RSpec helper method to show rendered views in browser (`save_and_open_page` does not work for rendered pages in view specs)
screen shot 2015-02-19 at 10 14 46 pm

### DIFF
--- a/spec/features/user_is_creating_multiple_sponsors_spec.rb
+++ b/spec/features/user_is_creating_multiple_sponsors_spec.rb
@@ -1,0 +1,51 @@
+require 'rails_helper'
+
+RSpec.feature 'User enters new sponsor data', :type => :feature do
+
+  background do
+    an_agent_exists
+    i_sign_in_as_admin
+  end
+
+  scenario 'User clicks the Create and Add Another button' do
+    when_i_enter_new_sponsor_info
+    and_i_click_create_and_add_another_button
+
+    then_the_sponsor_should_be_saved
+    and_i_should_be_on_new_sponsor_page
+  end
+end
+
+def an_agent_exists
+  FactoryGirl.create :agent
+end
+
+def i_sign_in_as_admin
+  admin = FactoryGirl.create :admin_user
+  visit new_admin_user_session_path
+  fill_in 'Email', with: admin.email
+  fill_in 'Password', with: admin.password
+  click_button 'Login'
+end
+
+def when_i_enter_new_sponsor_info
+  visit new_hq_sponsor_path
+  fill_in 'Name', with: 'C. Montgomery Burns'
+  fill_in 'Requested orphan count', with: '1'
+  select 'London', from: 'sponsor[branch_id]'
+  fill_in 'New city name', with: 'London'
+  select User.first.user_name, from: 'sponsor[agent_id]'
+end
+
+def and_i_click_create_and_add_another_button
+  click_button 'Create and Add Another'
+end
+
+def then_the_sponsor_should_be_saved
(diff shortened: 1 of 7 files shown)


spec/support/view_helpers.rb
```diff
@@ -0,0 +1,14 @@
...
+    File.open(filename, 'w') { |file| file.write(rendered) }
+    Launchy.open filename
+  rescue LoadError
+    warn 'You need to install launchy to open pages: `gem install launchy`'
+  end
+end
+
+RSpec.configure do |c|
+  c.include ViewHelpers, :type => :view
+end
```
razorcd: :+1:

spec/features/user_is_creating_multiple_sponsors_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+end
+
+def then_the_sponsor_should_be_saved
+  expect(page).to have_text 'Sponsor successfuly created'
+  expect(Sponsor.last.name).to eq 'C. Montgomery Burns'
+end
+
+def and_i_should_be_on_new_sponsor_page
+  expect(page).to have_text 'New Sponsor'
+end
```
razorcd: where do you think these methods definitions should go?  
`spec/features/steps` 
`spec/helpers/steps` 
?

NikitaAvvakumov: Not sure about the file name(s), I suppose whatever we agree on. The nice thing here vs. Cucumber is that only some steps can be defined in common files and thus be accessible from every feature, but those specific to a particular feature can continue to live right below the feature in the same file. 

I also propose that we keep features short and specific, whether we write them in RSpec or Cucumber - like here, it's not a generic "sponsors_spec.rb" that catches all sponsor-related tests, but a focused "user_is_creating_multiple_sponsors_spec.rb".

spec/features/user_is_creating_multiple_sponsors_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  click_button 'Create and Add Another'
+end
+
+def then_the_sponsor_should_be_saved
+  expect(page).to have_text 'Sponsor successfuly created'
+  expect(Sponsor.last.name).to eq 'C. Montgomery Burns'
+end
+
+def and_i_should_be_on_new_sponsor_page
+  expect(page).to have_text 'New Sponsor'
```
razorcd: isn't   `expect(current_path).to eq new_hq_sponsor_path`   a safer test for this?

spec/features/user_is_creating_multiple_sponsors_spec.rb
razorcd: agreed
and :shipit:

spec/features/user_is_creating_multiple_sponsors_spec.rb
NikitaAvvakumov: Definitely

gmunumel:
Looks nice to me :+1:

PR #336, opened 2015-02-21
OSRA380 - View and Controller for show action - Orphans

gmunumel:
OSRA380 - View and Controller for show action - Orphans

### DIFF
--- a/app/views/hq/orphans/show.html.haml
+++ b/app/views/hq/orphans/show.html.haml
@@ -36,10 +36,10 @@
         = @orphan.priority
 .panel.panel-default
   .panel-heading 
-    %h3.panel-title Parent's Details
+    %h3.panel-title Parents' Details
   .panel-body
     %dl.dl-horizontal
-      %dt Father Given Name
+      %dt Father's Given Name
       %dd
         = @orphan.father_given_name
     %dl.dl-horizontal
@@ -106,16 +106,20 @@
       %dt Street
       %dd
         = @orphan.current_address.street
-%dl.dl-horizontal
-  %dt Sponsored by Another Org
-  %dd
-    = @orphan.sponsored_by_another_org ? 'Yes' : 'No'
-%dl.dl-horizontal
-  %dt Minor Siblings Count
-  %dd
-    = @orphan.minor_siblings_count
-%dl.dl-horizontal
-  %dt Sponsored Minor Siblings Count
-  %dd
-    = @orphan.sponsored_minor_siblings_count
+.panel.panel-default
+  .panel-heading 
+    %h3.panel-title Additional Details
+  .panel-body
+    %dl.dl-horizontal
+      %dt Spnsr Another Organization
+      %dd
+        = @orphan.sponsored_by_another_org ? 'Yes' : 'No'
+    %dl.dl-horizontal
+      %dt Minor Siblings Count
+      %dd
+        = @orphan.minor_siblings_count
+    %dl.dl-horizontal
+      %dt Spnsr Minor Siblings Count
+      %dd
+        = @orphan.sponsored_minor_siblings_count
 
--- a/app/views/hq/orphans/show.html.haml
+++ b/app/views/hq/orphans/show.html.haml
@@ -1,0 +1,127 @@
+%div.action_items
+  %div.instance
+    = link_to 'Edit Orphan', edit_hq_orphan_path(@orphan.id),
+      class: 'btn btn-default', role: 'button'
+    = link_to 'Link to Orphan', '#',
+      class: 'btn btn-default', role: 'button'
+
+%dl.dl-horizontal
+  %dt Orphan Name
+  %dd
+    = @orphan.name
+%dl.dl-horizontal
+  %dt OSRA Num
+  %dd
+    = @orphan.osra_num
+%dl.dl-horizontal
+  %dt Father Given Name
+  %dd
+    = @orphan.father_given_name
+%dl.dl-horizontal
+  %dt Family Name
+  %dd
+    = @orphan.family_name
+%dl.dl-horizontal
+  %dt Father is Alive
+  %dd
+    = @orphan.father_alive ? 'Yes' : 'No'
+%dl.dl-horizontal
+  %dt Father is Martyr
+  %dd
+    = @orphan.father_is_martyr ? 'Yes' : 'No'
+%dl.dl-horizontal
+  %dt Father Date of Death
+  %dd
+    = format_full_date @orphan.father_date_of_death
+%dl.dl-horizontal
+  %dt Mother Name
+  %dd
+    = @orphan.mother_name
+%dl.dl-horizontal
+  %dt Mother is Alive
+  %dd
+    = @orphan.mother_alive ? 'Yes' : 'No'
+%dl.dl-horizontal
+  %dt Date of Birth
+  %dd
+    = format_full_date @orphan.date_of_birth
+%dl.dl-horizontal
+  %dt Gender
+  %dd
+    = @orphan.gender
+%dl.dl-horizontal
+  %dt Contact Number
+  %dd
+    = @orphan.contact_number
+%dl.dl-horizontal
+  %dt Sponsored by Another Org
+  %dd
+    = @orphan.sponsored_by_another_org ? 'Yes' : 'No'
+%dl.dl-horizontal
+  %dt Minor Siblings Count
+  %dd
+    = @orphan.minor_siblings_count
+%dl.dl-horizontal
+  %dt Sponsored Minor Siblings Count
+  %dd
+    = @orphan.sponsored_minor_siblings_count
+%dl.dl-horizontal
+  %dt Original Province Name
+  %dd
+    = @orphan.original_address.province.name
+%dl.dl-horizontal
+  %dt Original Province Code
+  %dd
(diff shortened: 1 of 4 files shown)


app/views/hq/orphans/show.html.haml
```diff
@@ -0,0 +1,127 @@
+%div.action_items
+  %div.instance
+    = link_to 'Edit Orphan', edit_hq_orphan_path(@orphan.id),
+      class: 'btn btn-default', role: 'button'
+    = link_to 'Link to Orphan', '#',
```
NikitaAvvakumov: In the ActiveAdmin interface, creation of sponsorships takes place only from a sponsor's show view. Let's keep it the same in HQ for now, and make the change if the client requests it.

app/views/hq/orphans/show.html.haml
```diff
@@ -0,0 +1,127 @@
...
+    = @orphan.minor_siblings_count
+%dl.dl-horizontal
+  %dt Sponsored Minor Siblings Count
+  %dd
+    = @orphan.sponsored_minor_siblings_count
+%dl.dl-horizontal
+  %dt Original Province Name
+  %dd
+    = @orphan.original_address.province.name
+%dl.dl-horizontal
```
NikitaAvvakumov: It seems that there is no need to display province codes here

app/views/hq/orphans/show.html.haml
```diff
@@ -0,0 +1,127 @@
+%div.action_items
+  %div.instance
+    = link_to 'Edit Orphan', edit_hq_orphan_path(@orphan.id),
+      class: 'btn btn-default', role: 'button'
+    = link_to 'Link to Orphan', '#',
+      class: 'btn btn-default', role: 'button'
+
+%dl.dl-horizontal
```
NikitaAvvakumov: Let's break this long display of attributes into thematic panels, the way it's currently done in the AA interface. Have a look at http://getbootstrap.com/components/#panels

app/views/hq/orphans/show.html.haml
```diff
@@ -0,0 +1,127 @@
...
+  %dt Minor Siblings Count
+  %dd
+    = @orphan.minor_siblings_count
+%dl.dl-horizontal
+  %dt Sponsored Minor Siblings Count
+  %dd
+    = @orphan.sponsored_minor_siblings_count
+%dl.dl-horizontal
+  %dt Original Province Name
```
NikitaAvvakumov: Once these fields are wrapped into 'Original Address' and 'Current Address' panels, the words 'Original' and 'Current' can be omitted from individual labels.

app/views/hq/orphans/show.html.haml
```diff
@@ -0,0 +1,127 @@
...
+    = @orphan.original_address.neighborhood
+%dl.dl-horizontal
+  %dt Original Street
+  %dd
+    = @orphan.original_address.street
+%dl.dl-horizontal
+  %dt Current Province Name
+  %dd
+    = @orphan.current_address.province.name
+%dl.dl-horizontal
```
NikitaAvvakumov: As above

app/views/hq/orphans/show.html.haml
```diff
@@ -0,0 +1,127 @@
...
+    = @orphan.current_address.neighborhood
+%dl.dl-horizontal
+  %dt Current Street
+  %dd
+    = @orphan.current_address.street
+%dl.dl-horizontal
+  %dt Name Status
+  %dd
+    = @orphan.orphan_status.name
+%dl.dl-horizontal
```
NikitaAvvakumov: Don't need the code

app/views/hq/orphans/show.html.haml
```diff
@@ -0,0 +1,127 @@
...
+    = @orphan.orphan_status.code
+%dl.dl-horizontal
+  %dt Priority
+  %dd
+    = @orphan.priority
+%dl.dl-horizontal
+  %dt Sponsorship Status
+  %dd
+    = @orphan.orphan_sponsorship_status.name
+%dl.dl-horizontal
```
NikitaAvvakumov: As above

app/views/hq/orphans/show.html.haml
```diff
@@ -0,0 +1,127 @@
...
+  %dt Current Neighborhood
+  %dd
+    = @orphan.current_address.neighborhood
+%dl.dl-horizontal
+  %dt Current Street
+  %dd
+    = @orphan.current_address.street
+%dl.dl-horizontal
+  %dt Name Status
```
NikitaAvvakumov: Just 'Status' or 'Orphan Status'. I would be more natural to display these attributes next to others that pertain directly to the orphan. See the current AA interface - the grouping of fields into panels there feels logical and is client-approved.

spec/controllers/hq/orphans_controller_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'rails_helper'
+require 'will_paginate/array'
+
+RSpec.describe Hq::OrphansController, type: :controller do
+    let(:orphan) {build_stubbed :orphan}
+    let(:orphans) {(1..5).map {build_stubbed :orphan}}
```
NikitaAvvakumov: Don't need this here until `#index` is implemented

spec/views/hq/orphans/show_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'rails_helper'
+
+RSpec.describe "hq/orphans/show.html.haml", type: :view do
+  let(:orphans) { FactoryGirl.build_stubbed(:orphan_list) }
```
NikitaAvvakumov: The orphan factory builds the `orphan_list` association. Do you need to create it here explicitly?

spec/views/hq/orphans/show_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  let(:orphan) { FactoryGirl.build_stubbed(:orphan, orphan_list: orphans) }
+
+  describe 'the orphan exists' do
+    before :each do
+      assign(:orphan, orphan)
```
NikitaAvvakumov: This is a minor style point: when you need to use a variable in the spec and also have it assigned to an instance variable in the view being tested, you can replace the lines
```ruby
let(:orphan) { build_stubbed :orphan }
before do
  assigns :orphan, orphan
  render
end
```
with
```ruby
before do
  @orphan = build_stubbed :orphan
  render
end
```
The `@orphan` variable is then used in both the view and the spec (you would need to change all references to `orphan` to `@orphan` in the `it` blocks).

gmunumel:
@NikitaAvvakumov Thanks for the clarification. I will try to made all these changes

spec/views/hq/orphans/show_spec.rb
gmunumel: Yes, you are right. I try to follow the same idea of `sponsor`. But for orphan is more simple.

app/views/hq/orphans/show.html.haml
gmunumel: I can group and try to mimic the look of AA interface, for instance, made a panel for orphan details, parent's details, etc. Is it ok?

NikitaAvvakumov: We don't need to follow the AA look verbatim, but with such a long list of attributes some kind of grouping into categories is needed. You can take AA as a starting point, but feel free to deviate from it if something doesn't feel right to you.

gmunumel:
Screenshot of the orphan right now.

screenshot from 2015-02-24 20 10 29

gmunumel:
Do I need to write behaviors test for this?

NikitaAvvakumov:
@gmunumel We haven't written integration tests for code like this so far, though I'm starting to wonder whether a very high-level feature can be created to test the entirety of the CRUD operations for a particular model - e.g.
```cucumber
Scenario: admin user manipulates orphan data
  Given an orphan record exists
  Then I should be able to look at it
  And I should be able to change it
  And I should be able to see it in a list of all orphan records
```
What do people think?

gmunumel:
I think sooner or later we will need them. :+1: But we can plan to add them in the next sprint.

app/views/hq/orphans/show.html.haml
```diff
@@ -0,0 +1,121 @@
...
+      %dt Sponsorship Status
+      %dd
+        = @orphan.orphan_sponsorship_status.name
+    %dl.dl-horizontal
+      %dt Priority
+      %dd
+        = @orphan.priority
+.panel.panel-default
+  .panel-heading 
+    %h3.panel-title Parent's Details
```
NikitaAvvakumov: "Parents' details"

app/views/hq/orphans/show.html.haml
```diff
@@ -0,0 +1,121 @@
...
+  .panel-body
+    %dl.dl-horizontal
+      %dt Father Given Name
```
NikitaAvvakumov: Let's take this opportunity to correct the app's grammar: "Father's Given Name" & "Mother's Given Name". This has bugged me for a long time :)

razorcd:
@NikitaAvvakumov : yes, for the CRUD operations we ca do that

app/views/hq/orphans/show.html.haml
gmunumel: sure:+1:

PR #337, opened 2015-02-23
Osra337 start dates not prior to osra establishment date

razorcd:
OSRA-337: All start dates should not precede OSRA establishment date of April 1, 2012
https://osraav.atlassian.net/browse/OSRA-377


app/models/partner.rb
```diff
@@ -30,6 +31,8 @@ def self.all_names
 
   private
 
+  include StartdateBeyondOsraEstablishmentValidation
+
```
razorcd: other possibility would be to also include the 
`validate :start_date_beyound_OSRA_establishment_date, if: :start_date`
in the module. I decided not to.

razorcd:
Form validation:
osra_establishment_date_validation

Sponsorship flash message:
osra_establishment_sponsorship_start_date_validation

app/models/concerns/startdate_beyond_osra_establishment_validation.rb
```diff
@@ -0,0 +1,12 @@
+module StartdateBeyondOsraEstablishmentValidation
+	include ValidatorHelpers
+
+	OSRA_ESTABLISHMENT_DATE = Date.new(2012,04,01)
```
gmunumel: Is there a file that Rails use to keep global variables? I think it is more tidy

razorcd: yes, we were thinking about it. There is also the priority countries for the select option, the pagination page limit, etc
I would like to have them all in one file.

gmunumel: Cool :ok:

NikitaAvvakumov: I'm going to disagree: `OSRA_ESTABLISHMENT_DATE` is a part of the business logic of our application, used by multiple entities. Something like this does belong in a central shared location. On the other hand, pagination params are specific not only to individual models, but are also relevant only to HTML rendering by Rails. If our app gets set up to e.g. also return JSON structures to web services or mobile clients, these constants lose all meaning. Country selection falls into the same category - it's something unique to Rails sponsor views. I'd prefer not to keep such constants mixed in with those that belong to the core functionality of OSRA.

app/models/concerns/startdate_beyond_osra_establishment_validation.rb
```diff
@@ -0,0 +1,12 @@
+module StartdateBeyondOsraEstablishmentValidation
+	include ValidatorHelpers
```
NikitaAvvakumov: Indentation got messed up here

app/models/partner.rb
NikitaAvvakumov: Agreed.

features/aa/aa_features/sponsorship.feature
```diff
@@ -49,6 +49,13 @@ Feature:
     When I click the "Sponsor this orphan" link for orphan "First Orphan"
     Then I should see "Start date can not be later than the first of next month"
 
+  Scenario: Sponsorship can not start later than OSRA establishment date
```
NikitaAvvakumov: ".. can not start _earlier_", right?

lib/validators/validator_helpers.rb
```diff
@@ -0,0 +1,13 @@
+module ValidatorHelpers
```
NikitaAvvakumov: Since this module is also used directly in models, would a different name be more appropriate? Simply `DateValidator`?

app/models/concerns/startdate_beyond_osra_establishment_validation.rb
gmunumel: But then at least add a formal documentation about them. I don't want to be looking for them in many files, that's the purpose of the global shared location.

app/models/validators/beyond_osra_establishment_validator.rb
```diff
@@ -0,0 +1,13 @@
+class BeyondOsraEstablishmentValidator < ActiveModel::EachValidator
+	OSRA_ESTABLISHMENT_DATE = Date.new(2012,04,01)
+	include Helpers
+
+	def validate_each(record, attribute, value)
+		if !valid_date? value
+			record.errors[attribute] << ('is not a valid date')
+		elsif value < OSRA_ESTABLISHMENT_DATE
+			record.errors[attribute] << ("is not valid (cannot be earlier than the OSRA establishment date of #{OSRA_ESTABLISHMENT_DATE})")
```
razorcd: notice: I changed the message.

app/models/validators/beyond_osra_establishment_validator.rb
```diff
@@ -0,0 +1,13 @@
...
+		end
+	end
```
razorcd: Maybe we should make separate validators for `valid_date` too? Either way, still have to check here `if !valid_date? value` first. (I decided not to)

app/models/validators/beyond_osra_establishment_validator.rb
```diff
@@ -0,0 +1,13 @@
+class BeyondOsraEstablishmentValidator < ActiveModel::EachValidator
+  OSRA_ESTABLISHMENT_DATE = Date.new(2012,04,01)
+  include Helpers
+
+  def validate_each(record, attribute, value)
+    if !valid_date? value
+      record.errors[attribute] << ('is not a valid date')
+    elsif value < OSRA_ESTABLISHMENT_DATE
+      record.errors[attribute] << ("is not valid (cannot be earlier than the OSRA establishment date of #{OSRA_ESTABLISHMENT_DATE})")
```
razorcd: notice: changed message

app/models/validators/beyond_osra_establishment_validator.rb
```diff
@@ -0,0 +1,13 @@
+class BeyondOsraEstablishmentValidator < ActiveModel::EachValidator
+  OSRA_ESTABLISHMENT_DATE = Date.new(2012,04,01)
+  include Helpers
+
+  def validate_each(record, attribute, value)
+    if !valid_date? value
+      record.errors[attribute] << ('is not a valid date')
```
razorcd: Maybe we should make separate validators for `valid_date` too? Either way, still have to check here `if !valid_date? value` first. (I decided not to)

razorcd:
UPDATED

model helpers are in `app/models/helpers/helpers.rb`
validators are in `app/models/validators/...`

gmunumel:
Good work :+1: 

I like the separation.

app/models/validators/beyond_osra_establishment_validator.rb
NikitaAvvakumov: Date format will be automatically validated since the value is stored as a `date` in the db.

razorcd: true. So we will keep this as it is?

app/models/helpers/helpers.rb
```diff
@@ -0,0 +1,13 @@
+module Helpers
```
NikitaAvvakumov: Can this go into app/models/concerns? 
Also, in the interest of keeping everything small, modular and single-responsibility, how about renaming this to e.g. 'date_validator.rb' instead of the generic 'helpers.rb' that may invite people to dump code here in the future?

NikitaAvvakumov:
There is also a `valid_date?` method in the Orphan model that needs to be taken out.

NikitaAvvakumov:
We have spec/lib/validators/date_not_in_future_validator_spec.rb that should be moved to spec/validators/. It can also serve as an example of a thorough validator spec - then the validator can be thoroughly tested there, and in the models that use it there can be only a simple test to make sure the module is included.

app/models/helpers/helpers.rb
razorcd: Yes, I see your point. 
I intentionally created `helpers.rb` so people to dump all helpers there. Otherwise we will have a bunch of `.rb` files with only one method in it and that will make rails load even slower.

app/models/validators/beyond_osra_establishment_validator.rb
NikitaAvvakumov: It can be reduced to
```ruby
if valid_date?(value) && (value < OSRA_ESTABLISHMENT_DATE
  record.errors[attribute] << ("is not valid (cannot be earlier than the OSRA establishment date of #{OSRA_ESTABLISHMENT_DATE})")
end
```

app/models/helpers/helpers.rb
NikitaAvvakumov: I'm all in favour of single-method modules (or classes). Mixing in a `Helpers` module with a dozen different unrelated methods into a model just because it needs e.g. date validation creates unnecessary bloat. It can also create difficult-to-track problems if the model tries to implement a method with the same name as a method in a module. Including a small single-purpose `DateValidator` module, on the other hand, minimizes such conflicts and makes it much clearer what method(s) a class inherits from the module.

gmunumel: In which cases is preferable to use helpers then?

NikitaAvvakumov: We can call the single-method modules Helpers, e.g. DateHelper, NameHelper etc. - that's just semantics. What I'm opposed to is a single Helpers module that ends up holding all accessory methods.

gmunumel: Got it. :+1:

spec/models/partner_spec.rb
```diff
@@ -17,10 +17,25 @@
...
+  context 'start_date validation' do
+    describe 'valid date' do
+      [Date.current, "10-10-2012"].each do |good_date_value|
+        it { is_expected.to allow_value(good_date_value).for :start_date }
+      end
+
+      [7, 'yes', true].each do |bad_date_value|
+        it { is_expected.to_not allow_value(bad_date_value).for :start_date }
+      end
+    end
```
razorcd: I am working on refactoring this part.

I am already testing good and bad dates when I test the `valid_date?` helper method. So I only need to test the presence of the `date_not_in_future` validator here.

I would like to see here something like:
`it { is_expected.to validate_date_not_in_future_of :start_date }` (same as: ``it { is_expected.to validate_presence_of :start_date }`) but I can't find an rspec method to directly test custom validators.
I'll keep searching for a solution.

Any help is welcome!

razorcd:
PR UPDATED

Odd test failure. It is all green locally.

NikitaAvvakumov:
Fixed by rebuilding on Semaphore.

app/models/concerns/date_helpers.rb
```diff
@@ -0,0 +1,13 @@
+module DateHelpers
+
+  private
+
+  def valid_date?(date)
+    Date.parse(date.to_s)
+  rescue ArgumentError
+    return false
+  else
```
NikitaAvvakumov: This seems redundant and a bit awkward. First, `Date.parse` will return a truthy value if the string encodes a valid date, so the explicit return of `true` is not necessary. Second, the logic flow is "do something -> if it failed, send failure message -> if it didn't, send success message". More natural, for me at least, would be "do something and send success message -> it that failed, send failure message".

razorcd: You mean like this?
```ruby
def valid_date?(date)
   return Date.parse(date.to_s)
 rescue ArgumentError
   return false
end
```

Or even better `Date.parse(date.to_s) && return true`  ?

NikitaAvvakumov: The first one, because you still need to rescue possible `ArgumentErrors`

spec/models/validators/Rspec_support_model_helpers_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+      validates :date_attr, valid_date_presence: true
+    end
+  end
+
+  subject { test_model.new }
+
+  describe "should have a functional 'have_validation' matcher" do
+    it { is_expected.to have_validation ValidDatePresenceValidator, :on => :date_attr }
+    it { is_expected.to_not have_validation ValidDatePresenceValidator, :on => :other_attr }
+  end
```
razorcd: I am thinking about using a simbol instead of validators class name.
So it will be:
`it { is_expected.to have_validation :valid_date_presence, :on => :date_attr }`

spec/models/validators/Rspec_support_model_helpers_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+end
```
razorcd: this file tests the `/spec/support/model_helpers.rb`
please review the name and path of this file.

spec/models/validators/Rspec_support_model_helpers_spec.rb
gmunumel: With the symbol looks better. But it's a choose of tastes rather than with functionality.

razorcd: That's true. We are also using symbols when we declare the validation but the validation is actually a class... so I am not sure which one is better for the tests.

NikitaAvvakumov: I prefer symbols

spec/models/validators/Rspec_support_model_helpers_spec.rb
gmunumel: Why the `Rspec` name in the file? What is the purposes of it?

spec/models/validators/date_beyond_osra_establishment_validator_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'rails_helper'
+
+describe DateBeyondOsraEstablishmentValidator do
```
theB264: `RSpec.describe` is generally better to use than `describe` for the first call to describe.  If we ever turn off the Rails monkey-patching to increase app performance with `config.expose_dsl_globally= false` then this will be needed. `RSpec.describe` does everything `describe` does, plus more, and (I think) it is more readable what is going on.

spec/models/validators/date_beyond_osra_establishment_validator_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      attr_accessor :date_attr
+
+      validates :date_attr, date_beyond_osra_establishment: true
+    end
+  end
+
+  subject { test_model.new }
+
+  it 'passes when attribute is a valid date, later than the ostra establishment date' do
```
theB264: typo

spec/models/validators/valid_date_presence_validator_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
+
+describe ValidDatePresenceValidator do
```
theB264: `RSpec.describe` is preferred if possible

spec/models/validators/Rspec_support_model_helpers_spec.rb
razorcd: because it is testing an `Rspec` matcher. It is not a method from the actual application.

spec/models/validators/date_beyond_osra_establishment_validator_spec.rb
razorcd: i agree. thanks. :+1:

razorcd:
PR Updated

spec/models/validators/Rspec_support_model_helpers_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+      it { is_expected.to have_validation :valid_date_presence, :on => :date_attr }
+      it { is_expected.to_not have_validation :valid_date_presence, :on => :other_attr }
+    end
+
+    describe "with options" do
+      it { is_expected.to have_validation :valid_date_presence, :on => :date_attr,
+                                          :options => [:allow_nil, :allow_blank] }
+    end
+  end
+end
```
razorcd: The test failure messages look like this:

`
  Failure/Error: it { is_expected.to have_validation :valid_date_presence, :on => :date_attr }
   expected that #<#<Class:0x0000000a346c00>:0x0000000afba860> would have validation ':valid_date_presence' on ':date_attr' field
`

`
   Failure/Error: it { is_expected.to have_validation :valid_date_presence, :on => :date_attr, :options => [:allow_nil] }
     expected that #<#<Class:0x0000000a0bd9c0>:0x0000000ad55b88> would have validation ':valid_date_presence' on ':date_attr' field with the options [:allow_nil]
`

razorcd: Still not sure if filename and file location is good here.

app/models/partner.rb
```diff
@@ -9,7 +9,9 @@ class Partner < ActiveRecord::Base
 
   validates :name, presence: true, uniqueness: true
   validates :province, presence: true
+  validates :start_date, valid_date_presence: true
   validates :start_date, date_not_in_future: true
+  validates :start_date, date_beyond_osra_establishment: true
```
NikitaAvvakumov: Just checking: do people prefer these separate validations to chaining?
```ruby
validates :start_date, valid_date_presence: true,
  date_not_in_future: true,
  date_beyond_osra_establishment: true
```

app/models/orphan.rb
```diff
@@ -5,6 +5,8 @@ class Orphan < ActiveRecord::Base
   VALID_GESTATION = 1
 
   include Initializer
+  include DateHelpers
```
NikitaAvvakumov: Please remove the `valid_date?` method on L192

spec/models/orphan_spec.rb
```diff
@@ -97,13 +94,13 @@
...
+      it { is_expected.to have_validation :date_not_in_future, :on => :father_date_of_death }
+
       specify 'father_date_of_death must be present and should be in the past' do
         orphan.father_alive = false
-        expect(orphan).to validate_presence_of :father_date_of_death
         expect(orphan).to allow_value(5.days.ago, Date.current).for :father_date_of_death
-        [7, 'yes', true, 5.days.from_now].each do |bad_date_value|
-          expect(orphan).to_not allow_value(bad_date_value).for :father_date_of_death
-        end
+        expect(orphan).to_not allow_value(5.days.from_now).for :father_date_of_death
```
NikitaAvvakumov: I remember that this is what we agreed on in the scrum, but looking at it now I'm having doubts about its usefulness. It's testing only one of the validations in a very limited way, while the validations are tested thoroughly in their respective spec files, and their inclusion is tested on L97-98 above. It seems to me that this line does not offer any real value. What do you think?

spec/models/partner_spec.rb
```diff
@@ -17,10 +17,18 @@
...
-  it { is_expected.not_to allow_value(Date.tomorrow).for :start_date }
-  [7, 'yes', true].each do |bad_date_value|
-    it { is_expected.to_not allow_value(bad_date_value).for :start_date }
+  context 'start_date validation' do
+    describe 'valid date' do
+      it { is_expected.to have_validation :valid_date_presence, :on => :start_date }
+    end
+
+    describe 'not in future' do
+      it { is_expected.to_not allow_value("2011-04-04").for :start_date }
```
NikitaAvvakumov: As above, I think that replacing this with `it { is_expected.to have_validation :date_not_in_future, :on => :father_date_of_death }` is more appropriate. Do you think these three `start_date` validations can be done without the individual `describe` blocks?

spec/models/sponsor_spec.rb
```diff
@@ -29,13 +29,37 @@
...
+  context 'start_date validation' do
+    describe 'valid date' do
+      it { is_expected.to have_validation :valid_date_presence, :on => :start_date }
+    end
+
+    describe 'not beyond OSRA establishment date' do
+      it { is_expected.to have_validation :date_beyond_osra_establishment, :on => :start_date }
+    end
+
+    describe "date_not_beyond_first_of_next_month" do
```
NikitaAvvakumov: This looks like a very good next candidate for extraction into a validator so that this logic does not pollute the sponsor model & spec.

spec/models/sponsor_spec.rb
```diff
@@ -29,13 +29,37 @@
...
+        it { is_expected.to allow_value(good_date).for :start_date }
+      end
+
+      [second_of_next_month, two_months_ahead].each do |bad_date|
+        it { is_expected.to_not allow_value(bad_date).for :start_date }
+      end
+    end
+  end
+
   it { is_expected.to allow_value(nil, '', 'admin@example.com', 'some.email@192.168.100.100', 'grüner@grü.üne',
```
NikitaAvvakumov: This one too? I would be in favour of taking out everything that is not directly relevant to the core business logic of a model.

spec/models/sponsorship_spec.rb
```diff
@@ -42,16 +46,17 @@
...
-      it { is_expected.to_not allow_value("5-12").for :start_date }
-      it { is_expected.to_not allow_value(second_of_next_month).for :start_date }
-      it { is_expected.to_not allow_value(two_months_ahead).for :start_date }
+      it { is_expected.to have_validation :valid_date_presence, :on => :start_date }
 
-      it { is_expected.to allow_value(today).for :start_date }
-      it { is_expected.to allow_value(first_of_next_month).for :start_date }
-      it { is_expected.to allow_value(yesterday).for :start_date }
-      it { is_expected.to allow_value(last_day_of_the_month).for :start_date }
+      describe "start_date_no_later_than_1st_of_next_month validation" do
```
NikitaAvvakumov: Oh yeah, definitely let's extract this into a validator!

spec/models/validators/Rspec_support_model_helpers_spec.rb
NikitaAvvakumov: Would 'spec/support/model_helpers_spec.rb' be more appropriate?

NikitaAvvakumov: :+1:

spec/models/validators/date_beyond_osra_establishment_validator_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  subject { test_model.new }
+
+  it 'passes when attribute is a valid date, later than the OSRA establishment date' do
+    subject.date_attr = Date.current
+    expect(subject).to be_valid
+  end
+
+  it 'fails when attribute is a valid date, beyond OSRA establishment date' do
```
NikitaAvvakumov: "..valid date _prior to_ OSRA establishment date". Here and L26.

spec/models/validators/date_not_in_future_validator_spec.rb
```diff
@@ -13,23 +13,12 @@
...
-
-  it 'returns appropriate message when attribute is not a date' do
-    subject.date_attr = 'fake_date'
-    expect(subject).to_not be_valid
-    expect(subject.errors[:date_attr]).to eq ['is not a valid date']
-  end
-
-  it 'passes when the date attribute is in the past' do
+  it 'passes when the date attribute is today' do
     subject.date_attr = 4.days.ago
```
NikitaAvvakumov: Should this be `Date.current`?

spec/models/validators/valid_date_presence_validator_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      attr_accessor :date_attr, :other_attr
+
+      validates :date_attr, valid_date_presence: true
+    end
+  end
+
+  subject { test_model.new }
+
+  it 'passes when the date is valid' do
+    expect_any_instance_of(ValidDatePresenceValidator).to receive(:valid_date?).and_return true
```
NikitaAvvakumov: Not sure that this is a good approach here: we are testing that the validator works as expected, so stubbing out its logic doesn't make sense. I think we want something like date_beyond_osra_establishment_validator_spec.rb above.

spec/support/model_helpers.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  match do |model|
+    my_validators = model.class.validators.select { |v| v.class == validator_class and v.attributes.include? params[:on] }
+    my_validators = my_validators.select { |v| params[:options].all? { |po| v.options.keys.any? { |vo| vo.to_sym == po } } } if params[:options]
+    my_validators.size == 1
+  end
+
+  failure_message do |model|
+    message = "expected that #{model} would have validation ':#{validator}' on ':#{params[:on]}' field"
+    message = message + " with the options #{params[:options]}" if params[:options]
```
NikitaAvvakumov: It's best to avoid trailing qualifiers. i.e. prefer
```ruby
if params[:options]
  message << " with the options..."
end
```

NikitaAvvakumov: Prefer string concatenation to addition :arrow_up:

spec/support/model_helpers.rb
```diff
@@ -0,0 +1,26 @@
...
+  validator_class = eval(validator.to_s.camelize + "Validator")
+
+  match do |model|
+    my_validators = model.class.validators.select { |v| v.class == validator_class and v.attributes.include? params[:on] }
+    my_validators = my_validators.select { |v| params[:options].all? { |po| v.options.keys.any? { |vo| vo.to_sym == po } } } if params[:options]
+    my_validators.size == 1
+  end
+
+  failure_message do |model|
+    message = "expected that #{model} would have validation ':#{validator}' on ':#{params[:on]}' field"
```
NikitaAvvakumov: For conciseness, I suggest "expected #{model} to have validation..." & "expected #{model} NOT to have validation..."

spec/support/model_helpers.rb
```diff
@@ -0,0 +1,26 @@
+# use like this:
+# it { is_expected.to have_validation :valid_date_presence,
+#                                     :on => :start_date,
+#                                     :options => [:allow_blank, :allow_nil]
+#                                     }
+RSpec::Matchers.define :have_validation do |validator, params|
+  validator_class = eval(validator.to_s.camelize + "Validator")
+
+  match do |model|
+    my_validators = model.class.validators.select { |v| v.class == validator_class and v.attributes.include? params[:on] }
```
NikitaAvvakumov: Best to use `&&` instead of `and` other than in cases of control flow (as in `redirect_to some_path and return`). Since the order of precedence is different between `&&` & `and`, you'll probably need parentheses around `v.class == validator_class`.

spec/support/model_helpers.rb
```diff
@@ -0,0 +1,26 @@
...
+    my_validators = my_validators.select { |v| params[:options].all? { |po| v.options.keys.any? { |vo| vo.to_sym == po } } } if params[:options]
```
NikitaAvvakumov: I would really like to see this clarified. Here's my attempt, though I'm not sure how well this all will work in the context of an RSpec custom matcher.

Testing that a particular validator is only used once for a particular attribute is a good idea. Is this the best way to handle that?

The logic for options is as follows: if `params[:options]` are passed to the matcher, ensure that the validator in the model gets passed the same options, correct?
```ruby
match do |model|
  @validator = model.class.validators.select { |v| (v.class == validator_class) && v.attributes.include? params[:on] }
  validator_included_once? && validator_receives_correct_options?
end

def validator_included_once?
  @validator.size == 1
end

def validator_receives_correct_options?
  return true unless params[:options]
  params[:options].all? { |o| @validator.first.options.has_key?(o) }
end
```

app/models/partner.rb
razorcd: yes, chaining on multiple lines looks cleaner. I'll update the PR.

spec/support/model_helpers.rb
razorcd: Yes, I agree `&&` is more appropriate here.

spec/support/model_helpers.rb
razorcd: Yes, separating the logic for validator and options looks better. :+1:

spec/models/orphan_spec.rb
gmunumel: I think have it doesn't cause so much noise. Anyways, you are testing the wrong path, that I consider could be useful in this case.

app/models/concerns/date_helpers.rb
```diff
@@ -0,0 +1,14 @@
+module DateHelpers
+
+  private
+
+  def valid_date?(date)
+     Date.parse(date.to_s) and return true
```
christophermlne: this looks strange.

return true if Date.parse(date.to_s)  ?

spec/models/concerns/date_helpers_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+RSpec.describe DateHelpers do
+  let(:test_class) do
+    Class.new { include DateHelpers }
+  end
+
+  subject { test_class.new }
+
+  context 'valid_date?' do
+    it 'method should be defined' do
+      expect(subject.private_methods.include? :valid_date?).to eq true
```
christophermlne: why test this?

spec/models/concerns/date_helpers_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    it 'should return true for a valid date' do
+      [Date.current, "10-10-2012"].each do |good_date_value|
+        expect(subject.send(:valid_date?, good_date_value)).to eq true
```
christophermlne: why are we testing private methods?

spec/models/sponsor_spec.rb
```diff
@@ -29,13 +29,37 @@
...
+
+    describe 'not beyond OSRA establishment date' do
+      it { is_expected.to have_validation :date_beyond_osra_establishment, :on => :start_date }
+    end
+
+    describe "date_not_beyond_first_of_next_month" do
+      past = 1.year.ago
+      today = Date.current
+      first_of_next_month = today.beginning_of_month.next_month
+      yesterday = today.yesterday
```
christophermlne: use let statements?

spec/models/validators/Rspec_support_model_helpers_spec.rb
christophermlne: is this Class getting torn down after the test?

spec/models/validators/Rspec_support_model_helpers_spec.rb
```diff
@@ -0,0 +1,28 @@
+#testing   /spec/support/model_helpers.rb
+require 'rails_helper'
+
+RSpec.describe "Rspec matchers" do
+  let(:test_model) do
+    Class.new do
```
christophermlne: is this Class getting torn down after the test?

spec/models/validators/date_beyond_osra_establishment_validator_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'rails_helper'
+
+RSpec.describe DateBeyondOsraEstablishmentValidator do
+  let(:test_model) do
+    Class.new do
```
christophermlne: again, are we sure the Class gets torn down after each test?

app/models/concerns/date_helpers.rb
NikitaAvvakumov: @christophermlne your suggestion is close to what was there initially. If `Date.parse` succeeds, its return value is a date object, which is truthy. The statement you propose then becomes `return true if true`, which is clearly redundant. This can then be shortened to `return Date.parse` or, since Ruby automatically returns the result of the last executed statement, simply `Date.parse`, which is what we had in this module to begin with. Ultimately, `true and return true` is not substantially different or less redundant than `return true if true`.

@razorcd will correct me if I'm wrong, but his desire to rewrite this arose from the fact that `valid_date?` is a predicate method, and it would be good for it to explicitly return `true` or `false`.

It is possible to coerce any value to a boolean by prepending it with `!!`, e.g. `!!Date.parse('01-01-2015') # true` - would that be an improvement?

christophermlne: return true if seems more idiomatic ruby, but i guess it doesn't matter...

spec/models/sponsor_spec.rb
razorcd: I created a new taks for this:
Extract model validations into ActiveModel validators
https://osraav.atlassian.net/browse/OSRA-395

razorcd: Oh, scratch this ^. I deleted it.
I see you already created a task https://osraav.atlassian.net/browse/OSRA-393

Looking at the other custom validators I don't see any other one that should be extracted to a ActiveModel validator for reuse.
For example in `Sponsorship` model we have `validate :end_date_not_before_start_date`. This is specific only for sponsorships . Any reason to move something like this in an ActiveModel validator?

spec/models/validators/valid_date_presence_validator_spec.rb
razorcd: Good and bad date values were already tested  in `models/concerns/date_helpers_spec.rb`
By NOT mocking the `valid_date?` method here, all the good and bad dates will need to be tested again here too. That is not good unit testing.
What do you think?

app/models/concerns/date_helpers.rb
razorcd: `return true if Date.parse(date.to_s)` is good but we need to remember that: `if (something falsy)` returns `false`

`!!Date.parse(date.to_s) ` is good too. I've seen it used many times.

How about `Date.parse(date.to_s)  ? true : false`

spec/models/concerns/date_helpers_spec.rb
razorcd: just to be sure the module is written correctly and the right module was imported.

spec/models/concerns/date_helpers_spec.rb
razorcd: good and bad dates are tested directly on the method so we don't have to test good and bad dates everywhere this method is used. We just check if this method is called or not.

spec/models/sponsor_spec.rb
razorcd: Yes, I favor `let` too.
But these variables are not inside an `it` block, it is a different scope. I don't think they have access to variables defined with `let`.

Also this `date_not_beyond_first_of_next_month` is outside of this task.

spec/models/validators/Rspec_support_model_helpers_spec.rb
razorcd: As far as I know, it is.
it only exists while the `describe "Rspec matchers"` is running.

razorcd:
PR UPDATED

spec/models/validators/valid_date_presence_validator_spec.rb
NikitaAvvakumov: You are right. I forgot that this validator outsources `valid_date?` to a module.

spec/support/model_helpers.rb
```diff
@@ -0,0 +1,35 @@
...
+# it { is_expected.to have_validation :valid_date_presence,
+#                                     :on => :start_date,
+#                                     :options => [:allow_blank, :allow_nil]
+#                                     }
+RSpec::Matchers.define :have_validation do |validator, params|
+  validator_class = eval(validator.to_s.camelize + "Validator")
+
+  match do |model|
+    @_validators = model.class.validators.select { |v| (v.class == validator_class) && (v.attributes.include? params[:on]) }
+    validator_included_once? && (validator_receives_correct_options? params[:options])
```
NikitaAvvakumov: This looks good, great job!
Small question: Can `validator_receives_correct_options?` not access params directly, in which case passing `params[:options]` as an argument is unnecessary?

NikitaAvvakumov:
This looks ready :+1:

spec/support/model_helpers.rb
razorcd: Thanks! And thanks for the reviews!

Yeah, I know, it is strange but from inside `validator_receives_correct_options?` method there is only access to instance variables. 
There is no access to `params`, `validator_class` and not even `model`.

razorcd:
:shipit:

spec/models/validators/valid_date_presence_validator_spec.rb
razorcd: I think I should add a new test in each validator spec to make sure it is checking for an invalid date too (for `DateBeyondOsraEstablishmentValidator` and `DateNotInFutureValidator`). 
At the moment I am only doing it in `ValidDatePresenceValidator` (in `valid_date_presence_validator_spec.rb` spec) which checks for valid and invalid dates .

Like:
```ruby
  it "allows date attribute to be an invalid date format" do
    expect_any_instance_of(DateBeyondOsraEstablishmentValidator).to receive(:valid_date?).and_return false
    expect(subject).to allow_value("invalid_date_format").for :date_attr
  end
#same for: DateNotInFutureValidator
```

In the model,  to block invalid dates, `validates :start_date, valid_date_presence: true` is needed.

gmunumel:
Yes, let it free :+1:

spec/support/model_helpers.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def validator_receives_correct_options? expected_options
+    return true unless expected_options
+
+    validators_options = @_validators.first.options
+    expected_options.all? do |k,v|
+      validators_options.has_key?(k) && validators_options[k] == v
+    end
+  end
+end
```
razorcd: notice change: 
` :options => [allow_nil, allow_blank]`
to
` :options => {allow_nil: true, allow_blank: true, if: :date_attr}`
*it also accepts `if/unless` options

Also added possibility to check only a few of the options.

razorcd:
PR UPDATED and rebased

PR #340, opened 2015-02-25
Minor refactoring and reorganizing of the Orphan model

NikitaAvvakumov:
This is minor code clean up in preparation for what will be a more substantial and meaningful refactoring. Note that this PR purposely avoids touching code that will be removed or changed when #325 and #337 get merged in. _It is probably a good idea to postpone merging this until #325 since rebasing this PR should be far less painful._

 - reduce the public API by making all validation methods private
 - extract previously hard-coded limits of "22 years of age" and "born 1 year after father's death" into constants, and refactor related validations
 - move checks for valid dates out of validation methods and into the `validate` calls - good idea?
 - unify setting of default attributes into a single parent method
 - extract several methods to better adhere to the Single Responsibility Principle
 - generally reorganize code into logical groups

### DIFF
--- a/app/models/orphan.rb
+++ b/app/models/orphan.rb
@@ -1,11 +1,12 @@
 class Orphan < ActiveRecord::Base
 
   NEW_SPONSORSHIP_SORT_SQL = '"orphan_sponsorship_statuses"."name", "orphans"."priority"  ASC'
+  AGE_OF_ELIGIBILITY_TO_JOIN = 22
+  VALID_GESTATION = 1
 
   include Initializer
-  after_initialize :default_orphan_status_active,
-                   :default_sponsorship_status_unsponsored,
-                   :default_priority_to_normal
+  after_initialize :set_defaults
+
   before_update :qualify_for_sponsorship_by_status,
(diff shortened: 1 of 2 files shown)


app/models/orphan.rb
```diff
@@ -132,46 +119,89 @@ def sponsorship_changed!
...
+
+  def generate_osra_num
+    self.osra_num = "#{province_code}%05d" % sequential_id
   end
 
-  def default_orphan_status_active
-    if new_record?
-      self.orphan_status ||= OrphanStatus.find_by_name 'Active'
+  def sponsored_siblings_does_not_exceed_siblings_count
+    if sponsored_siblings_exceed_total?
```
gmunumel: Why is this a function? Only has one line comparison and it never gets called in other part of the code?

razorcd: It separates a different behavior. It is easier to read, test and reuse. 
Also respects SOFA design principles where O says that methods should do only One thing.

gmunumel: I remember that one :+1:

PR #342, opened 2015-02-27
osra392 Add "Sponsored Orphans" list to the sponsor show view

razorcd:
Add "Sponsored Orphans" list to the sponsor show view
https://osraav.atlassian.net/browse/OSRA-392

- added 2 tables to the Sponsor#show view:
     - currently sponsored orphans (using a partial)
     - previously sponsored orphans (using a partial)
- added sponsorship routes for inactivate and destroy
- added Sponsorship#inactivate  controller action
- added Sponsorship#destroy  controller action
- added controller and view specs

Did not add integration tests.

### DIFF
--- a/app/controllers/hq/sponsors_controller.rb
+++ b/app/controllers/hq/sponsors_controller.rb
@@ -49,7 +49,7 @@ private
   end
 
   def load_active_and_inactive_sponsorships
-    sponsorships = Sponsorship.load_with_orphans_for_sponsor @sponsor.id
+    sponsorships = Sponsorship.load_with_orphans_for @sponsor
     @sponsorships_active = sponsorships.select {|sp| sp.active == true}
     @sponsorships_inactive = sponsorships.select {|sp| sp.active == false}
   end
--- a/app/controllers/hq/sponsors_controller.rb
+++ b/app/controllers/hq/sponsors_controller.rb
@@ -49,7 +49,7 @@ private
   end
 
   def load_active_and_inactive_sponsorships
-    sponsorships = Sponsorship.where("sponsor_id = ?", @sponsor.id).eager_load(:orphan).all
+    sponsorships = Sponsorship.load_with_orphans_for_sponsor @sponsor.id
     @sponsorships_active = sponsorships.select {|sp| sp.active == true}
     @sponsorships_inactive = sponsorships.select {|sp| sp.active == false}
   end
--- a/app/controllers/hq/sponsors_controller.rb
+++ b/app/controllers/hq/sponsors_controller.rb
@@ -49,7 +49,7 @@ private
   end
 
   def load_active_and_inactive_sponsorships
-    sponsorships = Sponsorship.where("sponsor_id = ?", @sponsor.id).eager_load(:orphan, :sponsor).all
+    sponsorships = Sponsorship.where("sponsor_id = ?", @sponsor.id).eager_load(:orphan).all
     @sponsorships_active = sponsorships.select {|sp| sp.active == true}
     @sponsorships_inactive = sponsorships.select {|sp| sp.active == false}
   end
--- a/app/controllers/hq/sponsors_controller.rb
+++ b/app/controllers/hq/sponsors_controller.rb
@@ -4,9 +4,8 @@ class Hq::SponsorsController < HqController
   end
 
   def show
-    @sponsor = Sponsor.find(params[:id])   #.includes(:sponsorships)
-    @sponsorships_active = @sponsor.sponsorships.select {|sp| sp.active == true}
-    @sponsorships_inactive = @sponsor.sponsorships.select {|sp| sp.active == false}
+    load_sponsor
+    load_active_and_inactive_sponsorships
   end
 
   def new
@@ -47,6 +46,12 @@ private
 
   def load_sponsor
     @sponsor= Sponsor.find(params[:id])
+  end
+
+  def load_active_and_inactive_sponsorships
+    sponsorships = Sponsorship.where("sponsor_id = ?", @sponsor.id).eager_load(:orphan, :sponsor).all
+    @sponsorships_active = sponsorships.select {|sp| sp.active == true}
+    @sponsorships_inactive = sponsorships.select {|sp| sp.active == false}
   end
 
   def save_sponsor
--- a/app/controllers/hq/sponsors_controller.rb
+++ b/app/controllers/hq/sponsors_controller.rb
@@ -4,7 +4,9 @@ class Hq::SponsorsController < HqController
   end
 
   def show
-    @sponsor = Sponsor.find(params[:id])
+    @sponsor = Sponsor.find(params[:id])   #.includes(:sponsorships)
+    @sponsorships_active = @sponsor.sponsorships.select {|sp| sp.active == true}
+    @sponsorships_inactive = @sponsor.sponsorships.select {|sp| sp.active == false}
   end
 
   def new
--- a/app/controllers/hq/sponsors_controller.rb
+++ b/app/controllers/hq/sponsors_controller.rb
@@ -49,7 +49,7 @@ private
   end
 
   def load_active_and_inactive_sponsorships
-    sponsorships = Sponsorship.load_with_orphans_for @sponsor
+    sponsorships = Sponsorship.where(sponsor: @sponsor)
     @sponsorships_active = sponsorships.select {|sp| sp.active == true}
     @sponsorships_inactive = sponsorships.select {|sp| sp.active == false}
   end
--- a/app/controllers/hq/sponsorships_controller.rb
+++ b/app/controllers/hq/sponsorships_controller.rb
@@ -2,18 +2,20 @@ class Hq::SponsorshipsController < HqController
 
   def inactivate
     sponsorship = Sponsorship.find(params[:id])
+    sponsor = sponsorship.sponsor
 
     @sponsorship_inactivator = InactivateSponsorship.new(sponsorship: sponsorship,
                                                          end_date: params[:sponsorship][:end_date])
-    inactivate_sponsorship and redirect_to hq_sponsor_path(sponsorship.sponsor_id)
+    inactivate_sponsorship and redirect_to hq_sponsor_path(sponsor)
   end
 
   def destroy
     sponsorship = Sponsorship.find(params[:id])
+    sponsor = sponsorship.sponsor
 
     @sponsorship_destructor = DestroySponsorship.new(sponsorship)
 
-    destroy_sponsorship and redirect_to hq_sponsor_path(sponsorship.sponsor_id)
+    destroy_sponsorship and redirect_to hq_sponsor_path(sponsor)
   end
 
   private
--- a/app/controllers/hq/sponsorships_controller.rb
+++ b/app/controllers/hq/sponsorships_controller.rb
@@ -2,22 +2,36 @@ class Hq::SponsorshipsController < HqController
 
   def inactivate
     sponsorship = Sponsorship.find(params[:id])
-    if sponsorship.inactivate params[:sponsorship][:end_date]
-      flash[:success] = 'Sponsorship link was successfully terminated'
-    else
-      flash[:error] = sponsorship.errors.full_messages || 'Sponsorship not terminated'
-    end
-    redirect_to hq_sponsor_path(sponsorship.sponsor_id)
+
+    @sponsorship_inactivator = InactivateSponsorship.new(sponsorship: sponsorship,
+                                                         end_date: params[:sponsorship][:end_date])
+    inactivate_sponsorship and redirect_to hq_sponsor_path(sponsorship.sponsor_id)
   end
 
   def destroy
     sponsorship = Sponsorship.find(params[:id])
-    if sponsorship.destroy
-      flash[:success] = 'Sponsorship was successfully deleted'
+
+    @sponsorship_destructor = DestroySponsorship.new(sponsorship)
+
+    destroy_sponsorship and redirect_to hq_sponsor_path(sponsorship.sponsor_id)
+  end
+
+  private
+
+  def inactivate_sponsorship
+    if @sponsorship_inactivator.call
+      flash[:success] = 'Sponsorship link was successfully terminated.'
     else
-      flash[:error] = 'Sponsorship not deleted'
+      flash[:error] = @sponsorship_inactivator.error_msg
     end
-    redirect_to hq_sponsor_path(sponsorship.sponsor_id)
+  end
+
+  def destroy_sponsorship
+    if @sponsorship_destructor.call
+      flash[:success] = 'Sponsorship record was successfully destroyed.'
+    else
+      flash[:error] = @sponsorship_destructor.error_msg
+    end
   end
 
 end
--- a/app/controllers/hq/sponsorships_controller.rb
+++ b/app/controllers/hq/sponsorships_controller.rb
@@ -1,0 +1,23 @@
+class Hq::SponsorshipsController < HqController
+
+  def inactivate
+    sponsorship = Sponsorship.find(params[:id])
+    if sponsorship.inactivate params[:sponsorship][:end_date]
+      flash[:success] = 'Sponsorship link was successfully terminated'
+    else
+      flash[:error] = sponsorship.errors.full_messages || 'Sponsorship not terminated'
+    end
+    redirect_to hq_sponsor_path(sponsorship.sponsor_id)
+  end
+
+  def destroy
+    sponsorship = Sponsorship.find(params[:id])
+    if sponsorship.destroy
+      flash[:success] = 'Sponsorship was successfully deleted'
+    else
+      flash[:error] = 'Sponsorship not deleted'
+    end
+    redirect_to hq_sponsor_path(sponsorship.sponsor_id)
+  end
+
+end
--- a/app/models/sponsorship.rb
+++ b/app/models/sponsorship.rb
@@ -28,6 +28,8 @@ class Sponsorship < ActiveRecord::Base
 
   scope :all_active, -> { where(active: true) }
   scope :all_inactive, -> { where(active: false) }
+  scope :load_with_orphans_for_sponsor,
+         ->(sponsor_id) {where("sponsor_id = ?", sponsor_id).eager_load(:orphan).all}
 
 private
 
--- a/app/models/sponsorship.rb
+++ b/app/models/sponsorship.rb
@@ -28,8 +28,8 @@ class Sponsorship < ActiveRecord::Base
 
   scope :all_active, -> { where(active: true) }
   scope :all_inactive, -> { where(active: false) }
-  scope :load_with_orphans_for,
-        ->(sponsor) {where(sponsor: sponsor).eager_load(:orphan).all}
+  
+  default_scope { includes(:sponsor, :orphan) }
 
 private
 
--- a/app/models/sponsorship.rb
+++ b/app/models/sponsorship.rb
@@ -28,8 +28,8 @@ class Sponsorship < ActiveRecord::Base
 
   scope :all_active, -> { where(active: true) }
   scope :all_inactive, -> { where(active: false) }
-  scope :load_with_orphans_for_sponsor,
-         ->(sponsor_id) {where("sponsor_id = ?", sponsor_id).eager_load(:orphan).all}
+  scope :load_with_orphans_for,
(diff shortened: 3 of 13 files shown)


razorcd:
UPDATED
Sponsorships are eager loading the associated orphans and sponsor.

app/controllers/hq/sponsors_controller.rb
```diff
@@ -47,6 +48,12 @@ def load_sponsor
     @sponsor= Sponsor.find(params[:id])
   end
 
+  def load_active_and_inactive_sponsorships
+    sponsorships = Sponsorship.where("sponsor_id = ?", @sponsor.id).eager_load(:orphan, :sponsor).all
```
razorcd: or we can use:
`sponsorships = Sponsorship.joins(:sponsor, :orphan).where("sponsor_id = ?",@sponsor.id).select("sponsorships.*, orphans.name as orphans_name, orphans.date_of_birth as orphans_date_of_birth, orphans.gender as orphans_gender")`

gmunumel: My only concern with the `eager` approach is with time consumption. But it looks nice and simpler than the join approach. If you are interested [this](http://blog.arkency.com/2013/12/rails4-preloading/) is a good article. Nice work cristian :+1:

NikitaAvvakumov: Is it necessary to eager_load sponsor here considering that the sponsor is already in memory as `@sponsor`? Also, what do you think about defining a default scope on Sponsorship that includes orphans & sponsors tables, the way @PurityControl has done for Orphan?

app/controllers/hq/sponsorships_controller.rb
```diff
@@ -0,0 +1,23 @@
+class Hq::SponsorshipsController < HqController
+
+  def inactivate
+    sponsorship = Sponsorship.find(params[:id])
+    if sponsorship.inactivate params[:sponsorship][:end_date]
```
NikitaAvvakumov: Sorry, all this needs changing now that service objects PR has been merged in. On the plus side, I tried to make the AA sponsorships controller code as portable as possible, so you should be able to mostly copy-paste it in here.

app/controllers/hq/sponsors_controller.rb
razorcd: eager loading the `:sponsor` too was necessary because the `sponsorship.sponsor_id` delegated to `sponsor.id`. But since that was fixed the `:sponsor` is not needed here.
Good find! Thanks :+1:

razorcd:
PR Updated

app/controllers/hq/sponsors_controller.rb
```diff
@@ -47,6 +48,12 @@ def load_sponsor
     @sponsor= Sponsor.find(params[:id])
   end
 
+  def load_active_and_inactive_sponsorships
+    sponsorships = Sponsorship.load_with_orphans_for_sponsor @sponsor.id
+    @sponsorships_active = sponsorships.select {|sp| sp.active == true}
+    @sponsorships_inactive = sponsorships.select {|sp| sp.active == false}
+  end
```
razorcd: notice: i added the sponsorship scope

app/controllers/hq/sponsorships_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    inactivate_sponsorship and redirect_to hq_sponsor_path(sponsorship.sponsor_id)
+  end
+
+  def destroy
+    sponsorship = Sponsorship.find(params[:id])
+
+    @sponsorship_destructor = DestroySponsorship.new(sponsorship)
+
+    destroy_sponsorship and redirect_to hq_sponsor_path(sponsorship.sponsor_id)
+  end
```
razorcd: notice the sponsor is not loaded anymore
this:
 ```
 sponsor = Sponsor.find(params[:sponsor_id])
 sponsorship = sponsor.sponsorships.find(params[:id])
```
was replaced with
`sponsorship = Sponsorship.find(params[:id])`

and redirecting to `sponsorship.sponsor_id`

Comment (gmunumel) [inline app/controllers/hq/sponsorships_controller.rb]: That's great :+1:

Comment (theB264) [inline app/controllers/hq/sponsorships_controller.rb]: Yeah I second that @gmunumel @razorcd .  I had long wondered why the `SponsorshipsController` was not using the `sponsorship_id`.  I think @nik mentioned it before as well.  This was long overdue

File: app/controllers/hq/sponsorships_controller.rb [hunk 5]
```diff
@@ -0,0 +1,37 @@
+class Hq::SponsorshipsController < HqController
+
+  def inactivate
+    sponsorship = Sponsorship.find(params[:id])
+
+    @sponsorship_inactivator = InactivateSponsorship.new(sponsorship: sponsorship,
+                                                         end_date: params[:sponsorship][:end_date])
+    inactivate_sponsorship and redirect_to hq_sponsor_path(sponsorship.sponsor_id)
```
NikitaAvvakumov: Why in so many places we are hard-coding the `id` as the identifier? Rails is capable of handling `redirect_to orphan_path(@orphan)` or even `redirect_to @orphan` (http://guides.rubyonrails.org/routing.html#creating-paths-and-urls-from-objects). If we e.g. implement vanity URLs and override `to_param`, we'll have to go through all this code and convert `sponsorship.sponsor.id` to `sponsorship.sponsor.slug` or such.

app/controllers/hq/sponsorships_controller.rb
NikitaAvvakumov: :+1: 
To complement this, it would be good to use [shallow nesting](http://guides.rubyonrails.org/routing.html#shallow-nesting) in routes.rb - collection actions remain scoped under the parent resource (here, sponsor), but member actions are not.

(Also, purely conceptually, destroying a sponsorship and then referencing it on L16 looks weird.)

app/models/sponsorship.rb
```diff
@@ -21,11 +21,13 @@ class Sponsorship < ActiveRecord::Base
...
   belongs_to :orphan
 
-  delegate :name, :additional_info, :id, to: :sponsor, prefix: true
+  delegate :name, :additional_info, to: :sponsor, prefix: true
   delegate :date_of_birth, :gender, to: :orphan, prefix: true
 
   scope :all_active, -> { where(active: true) }
   scope :all_inactive, -> { where(active: false) }
+  scope :load_with_orphans_for_sponsor,
+         ->(sponsor_id) {where("sponsor_id = ?", sponsor_id).eager_load(:orphan).all}
```
NikitaAvvakumov: I wonder whether `load_with_orphans_for @sponsor` would read better.
Also, a question that has come up before: Should the default scope for sponsorship eager load sponsor and orphan instead?

app/views/hq/sponsors/_sponsorships_active.html.haml
```diff
@@ -0,0 +1,47 @@
...
+              = link_to sponsorship.orphan.name, hq_orphan_path(sponsorship.orphan)
+            %td
+              = sponsorship.orphan.date_of_birth
+            %td
+              = sponsorship.orphan.gender
+            %td
+              = sponsorship.start_date
+            %td
+              =form_for(sponsorship,
+                  :url => {controller: :sponsorships, action: :inactivate, sponsor_id: sponsorship.sponsor_id, id: sponsorship.id},
```
NikitaAvvakumov: Shallow nesting will stop the need to provide the unused `sponsor_id`

config/environments/development.rb
```diff
@@ -26,4 +26,7 @@
   # This option may cause significant delays in view rendering with a large
   # number of complex assets.
   config.assets.debug = true
+
+  #disable the default querry caching
+  # config.middleware.delete "ActiveRecord::QueryCache"
```
NikitaAvvakumov: Please remove commented code

spec/controllers/hq/sponsors_controller_spec.rb
```diff
@@ -2,7 +2,9 @@
 require 'will_paginate/array'
 
 RSpec.describe Hq::SponsorsController, type: :controller do
-    let(:sponsor) {build_stubbed :sponsor}
+    let(:sponsorships_active) {(1..3).map {build_stubbed :sponsorship, active: true}}
```
NikitaAvvakumov:
```ruby
let(:active_sponsorships) { build_stubbed_list :sponsorship, 3, active: true }
```
gives the same result but seems more direct in its intentions.

spec/controllers/hq/sponsors_controller_spec.rb
```diff
@@ -19,8 +21,12 @@
 
   specify '#show' do
     expect(Sponsor).to receive(:find).and_return(sponsor)
+    expect(Sponsorship).to receive_message_chain(:where, :eager_load,:all).
```
NikitaAvvakumov: Missing space before `:all` :)

spec/controllers/hq/sponsorships_controller_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  before :each do
+    sign_in instance_double(AdminUser)
+  end
+
+  specify "#inactivate" do
+    expect(Sponsorship).to receive(:find).and_return(sponsorship)
+    put :inactivate, id: sponsorship.id, sponsor_id: sponsorship.sponsor_id, sponsorship: {end_date: Date.current}
+
+    expect(sponsorship.active).to be false
```
NikitaAvvakumov: I think the `#destroy` test below captures the essence of a controller spec better. The sponsorship's inactivation is performed by a service object, not the controller, which makes it a good candidate for stubbing just as `.destroy`. The controller only sends relevant instructions and decides what to do once it gets the response - whether the receivers were actually able to carry out those instructions is not its concern.

On the other hand, testing the unhappy paths of inactivation and destruction would be appropriate, as would testing that correct flash messages get set.

spec/views/hq/sponsors/_sponsorships_active_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'rails_helper'
+
+RSpec.describe 'hq/sponsors/_sponsorships_active.html.haml', type: :view do
+  let(:orphan) {build_stubbed :orphan}
+  let(:sponsorships_active) {(1..3).map {build_stubbed :sponsorship, active: true, orphan: orphan}}
```
NikitaAvvakumov: See my comment about `build_stubbed_list` above.

NikitaAvvakumov: You are potentially setting yourself up for trouble here - an orphan is only allowed to have 1 active sponsorship. If validations get triggered anywhere down the line, things will break for reasons that have nothing to do with this test. While that is unlikely in a view spec, I'd still prefer not to knowingly create invalid data.

Looking at the following tests, I don't think reusing a single orphan object buys anything. L4 can be simply deleted and reference to it from L5 removed.

spec/views/hq/sponsors/_sponsorships_active_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      render :partial => 'hq/sponsors/sponsorships_active.html.haml', :locals => {sponsorships: sponsorships_active}
+    end
+
+    it "should show table title" do
+      expect(rendered).to have_text "3 currently sponsored orphans"
+    end
+
+    it "should show sponsorships and linked orphans details" do
+      sponsorships_active.each do |sa|
+        expect(rendered).to have_link(sa.orphan.name, hq_orphan_path(sa.orphan_id))
```
NikitaAvvakumov: Again, prefer not to hard-code `id` as the param

app/views/hq/sponsors/_sponsorships_active.html.haml
```diff
@@ -0,0 +1,47 @@
...
+                  method: :put) do |f|
+                =f.submit "End Sponsorship", class: "btn btn-default"
+                =f.label :end_date, "on"
+                =f.text_field :end_date, value: Date.current
+            %td
+              =link_to "Delete",
```
NikitaAvvakumov: I originally had it as 'X' because of a). space; and b). sponsorship deletion is something that should not be a part of the standard workflow. The client requested it as an option for when they mess up sponsorship creation. So, I wanted deletion to be obscure, dangerous-looking and maybe convoluted somehow, to really send the message _"Are you sure?!!!"_.

If we decide to go the 'X' route, we can use Bootstrap's 'glyphicon-remove' coloured blood-red, poisonous-orange or something. @christophermlne, does that make you cringe? :)

spec/views/hq/sponsors/_sponsorships_active_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    it "should have Delete button" do
+      sponsorships_active.each do |sa|
+        expect(rendered).to have_link "Delete", hq_sponsor_sponsorship_path(sa.sponsor_id,sa)
+      end
+    end
+  end
+
+  describe "active sponsorships does not exist" do
```
NikitaAvvakumov: "do not"

spec/views/hq/sponsors/_sponsorships_inactive_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'rails_helper'
+
+RSpec.describe 'hq/sponsors/_sponsorships_inactive.html.haml', type: :view do
+  let(:orphan) {build_stubbed :orphan}
+  let(:sponsorships_inactive) {(1..2).map {build_stubbed :sponsorship, active: false, orphan: orphan}}
```
NikitaAvvakumov: While there is no conflict with validations, I'd still prefer to not reuse the orphan. It only saves creation of some in-memory objects, but makes the test on L16 less robust since only one sponsorship needs to have been rendered for it to pass.

spec/views/hq/sponsors/_sponsorships_inactive_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        expect(rendered).to have_link(sa.orphan.name, hq_orphan_path(sa.orphan_id))
+        expect(rendered).to have_text(sa.orphan.date_of_birth)
+        expect(rendered).to have_text(sa.orphan.gender)
+        expect(rendered).to have_text(sa.start_date)
+        expect(rendered).to have_text(sa.end_date)
+      end
+    end
+  end
+
+  describe "active sponsorships exist" do
```
NikitaAvvakumov: "inactive sponsorships do not exist"

app/views/hq/sponsors/_sponsorships_active.html.haml
gmunumel: I don't know if the client would like the "obscure" option if they didn't request that. I think give them the "Are you sure?!!" message won't warranty any mistake they can made. So I would prefer to show some options to the client and they choose.

app/controllers/hq/sponsorships_controller.rb
razorcd: I see your point.
`sponsorship.sponsor_id`was used instead `sponsorship.sponsor` to skip an extra sql query.
I guess it is not worth the sacrifice?

spec/controllers/hq/sponsors_controller_spec.rb
razorcd: nice!  `build_stubbed_list` is better. Thanks!

spec/controllers/hq/sponsorships_controller_spec.rb
razorcd: if we do `expect(sponsorship).to receive(:update!).and_return(true)` it will only return `true`, it will not change the `sponsorship.active`.   
Notice it is the same story we had with mocking the `.save`

spec/views/hq/sponsors/_sponsorships_active_spec.rb
razorcd: yes, good point! :+1:

config/environments/development.rb
```diff
@@ -26,4 +26,7 @@
   # This option may cause significant delays in view rendering with a large
   # number of complex assets.
   config.assets.debug = true
+
+  #disable the default query caching
+  # config.middleware.delete "ActiveRecord::QueryCache"
```
razorcd: I prefer to keep this here to use it when we want to disable query caching.

razorcd:
Delete button is a small red X button now.
inactive_sponsorships

razorcd:
PR UPDATED

gmunumel:
I wonder, why the date of "End Sponsorship" is available to edit? It shouldn't be just for show purposes?

razorcd:
Client can set the desired date when a sponsorship is ended. 
We also have validations on the that field. `Sponsorship#end_date`

spec/controllers/hq/sponsorships_controller_spec.rb
NikitaAvvakumov: Why is that a problem in the context of a controller spec? All we want to test is that the controller sends the right message, and then takes appropriate action based on response to that message. Just as we don't care that the sponsorship is actually destroyed below, so we don't care that the status actually changes - that is not the responsibility of the code under test.

app/models/sponsorship.rb
```diff
@@ -21,11 +21,13 @@ class Sponsorship < ActiveRecord::Base
...
   belongs_to :orphan
 
-  delegate :name, :additional_info, :id, to: :sponsor, prefix: true
+  delegate :name, :additional_info, to: :sponsor, prefix: true
   delegate :date_of_birth, :gender, to: :orphan, prefix: true
 
   scope :all_active, -> { where(active: true) }
   scope :all_inactive, -> { where(active: false) }
+  scope :load_with_orphans_for,
+        ->(sponsor) {where(sponsor: sponsor).eager_load(:orphan).all}
```
NikitaAvvakumov: @razorcd Any thoughts on eager loading orphan & sponsor in the default scope for sponsorship?

spec/controllers/hq/sponsorships_controller_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  before :each do
+    sign_in instance_double(AdminUser)
+  end
+
+  specify "#inactivate" do
+    expect(Sponsorship).to receive(:find).and_return(sponsorship)
+    put :inactivate, id: sponsorship.id, sponsorship: {end_date: Date.current}
+
+    expect(sponsorship.active).to be false
```
NikitaAvvakumov: I don't know if you'll see my other comment, so I'm just duplicating it here:
I'm still not convinced that there is any difference between the '#inactivate' and the '#destroy' specs, and am not sure why we can't stub `.update!` just as we do with `.destroy`.

razorcd: you can't stub `update!` because Rails `sponsorship.update!(active: false, end_date: end_date)` also changes `sponsorship.active` to `false`.

If you do:
```ruby
    expect(Sponsorship).to receive(:find).and_return(sponsorship)
    expect(sponsorship).to receive(:update!).and_return(true)
```
it will not change `sponsorship.active` to `false` and line 14 will fail.

And if you do:
```ruby
    expect(Sponsorship).to receive(:find).and_return(sponsorship)
    sponsorship.active = false
    expect(sponsorship).to receive(:update!).and_return(sponsorship)
```
then `:find` will also return a sponsorship with `sponsorship.active = false` and then the test has no value. ( because then we could change `sponsorship.update!(active: false, end_date: end_date)` to `sponsorship.update!(  end_date: end_date)` and the test would still pass)

NikitaAvvakumov: Sorry, I should have made myself clearer from the beginning: We should not be testing that the sponsorship is, in fact, inactivated, just as we don't test that the sponsorship actually gets destroyed below, i.e. L14 should not be here at all. Those are the responsibilities of the respective services that are exercised in their own specs. All we test here are the messages that the controller sends and its response to the replies that come back.

I am suggesting a spec along the lines of https://github.com/AgileVentures/osra/blob/develop/spec/controllers/admin/sponsorships_controller_spec.rb.

NikitaAvvakumov: Also note that without stubbing it is impossible to test the unhappy path - something that is still missing from this spec.

razorcd:
pr updated

We are mocking `:update!` and `:destroy!` but this is touching inside the services. It doesn't feel right but mocking the service itself would be a pain. 
Any thoughts on this?

razorcd:
updated

PR #343, opened 2015-03-06
OSRA-396 Replace orphan Statuses with enums

NikitaAvvakumov:
https://osraav.atlassian.net/browse/OSRA-396

Replace OrphanStatus & OrphanSponsorshipStatus models with enums `status` &
`sponsorship_status` in the Orphan model. Upshot: delete code & take advantage
of methods and scopes automatically implemented by Rails.


app/models/orphan.rb
```diff
@@ -1,14 +1,35 @@
...
+  ]
+
   AGE_OF_ELIGIBILITY_TO_JOIN = 22
   VALID_GESTATION = 1
 
+  SPONSORSHIP_STATUS_ORDERING = %Q{
+  CASE WHEN sponsorship_status = '#{Orphan.sponsorship_statuses[:previously_sponsored]}' THEN '1'
+       WHEN sponsorship_status = '#{Orphan.sponsorship_statuses[:unsponsored]}'          THEN '2'
+  END ASC
+  }
```
NikitaAvvakumov: Client requested ordering of eligible orphans as 'Previously Sponsored', 'Unsponsored'. Prior implementation relied on the fact that the former precedes the latter alphabetically. By contrast, this SQL explicitly specifies the order.

app/models/orphan.rb
```diff
@@ -69,18 +86,14 @@ class Orphan < ActiveRecord::Base
   accepts_nested_attributes_for :current_address, allow_destroy: true
   accepts_nested_attributes_for :original_address, allow_destroy: true
 
-  default_scope { includes(:partner, :orphan_sponsorship_status, :orphan_status, original_address: :province) }
+  default_scope { includes(:partner, original_address: :province) }
 
-  scope :active,
```
NikitaAvvakumov: `.active` scope is now automatically provided by Rails

app/models/orphan.rb
```diff
@@ -69,18 +86,14 @@ class Orphan < ActiveRecord::Base
...
-        -> { joins(:orphan_status).
-            where(orphan_statuses: { name: 'Active' }) }
   scope :currently_unsponsored,
-        -> { joins(:orphan_sponsorship_status).
-            where(orphan_sponsorship_statuses: { name: ['Unsponsored', 'Previously Sponsored'] }) }
+    -> { where(sponsorship_status: [ Orphan.sponsorship_statuses[:unsponsored],
+                                     Orphan.sponsorship_statuses[:previously_sponsored] ]) }
   scope :high_priority, -> { where(priority: 'High') }
-  scope :deep_joins, -> { joins(:orphan_sponsorship_status).joins(:original_address).joins(:partner) }
```
NikitaAvvakumov: No longer need the `.deep_joins` scope since the default scope includes same tables.

app/models/orphan.rb
```diff
@@ -96,16 +109,12 @@ def eligible_for_sponsorship?
     Orphan.active.currently_unsponsored.include? self
   end
 
-  def currently_sponsored?
-    sponsorships.all_active.present?
-  end
-
   def current_sponsorship
-    sponsorships.all_active.first if currently_sponsored?
+    sponsorships.all_active.first if sponsored?
```
NikitaAvvakumov: Relying on `sponsorship_status` to accurately represent current state instead of dynamically querying whether any active sponsorships are present. Good idea?

app/models/orphan.rb
```diff
@@ -175,16 +183,14 @@ def orphan_born_long_after_fathers_death?
   end
 
   def can_be_inactivated
-    if currently_sponsored?
-      errors.add :orphan_status,
+    if sponsored? && being_inactivated?
```
NikitaAvvakumov: I moved the `being_inactivated?` conditional here from the `validate` call on L74 as it makes the code more readable. Before, the method said "If sponsored, then add an error" and one would have to go back to the `validate` call to see that the method would only be called when an orphan is being inactivated. Now it says "If sponsored and being inactivated, then add an error", which can be understood in isolation. Thoughts?

db/schema.rb
```diff
@@ -11,18 +11,18 @@
...
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20141229224941) do
+ActiveRecord::Schema.define(version: 20150306182232) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
   create_table "active_admin_comments", force: :cascade do |t|
-    t.string   "namespace",     limit: 255
```
NikitaAvvakumov: I have no idea why the schema got string limits after the 4.2 upgrade, or why they've been taken out by resetting the db. The only relevant changes here are on old L148-149 and new L156-157.

features/aa/aa_features/orphans.feature
```diff
@@ -5,9 +5,9 @@ Feature:
 
   Background:
     Given the following orphans exist:
-      | name     | father_given_name | family_name | spon_status | death_date | mother   | birth_date | contact   | o_city  | o_province | o_hood  | c_city  | c_province | c_hood  |
```
NikitaAvvakumov: Sponsorship status is managed by the application and should never be specified explicitly, otherwise it creates logical inconsistencies. e.g. status is `sponsored` but there is no `current_sponsor`, which breaks the views.

spec/models/orphan_spec.rb
```diff
@@ -53,7 +44,7 @@
 
   it { is_expected.to_not allow_value(nil).for(:sponsored_by_another_org) }
 
-  #it { is_expected.to validate_numericality_of(:minor_siblings_count).only_integer.is_greater_than_or_equal_to(0) }
```
NikitaAvvakumov: I don't know why this was commented out. Seems to be a valid test.

spec/models/orphan_spec.rb
```diff
@@ -446,15 +425,6 @@
 
       describe 'scopes' do
 
-        specify '.active should correctly select active orphans only' do
```
NikitaAvvakumov: `.active` scope is implemented by Rails, no need to test

spec/models/orphan_spec.rb
```diff
@@ -464,23 +434,18 @@
                                                                     active_unsponsored_high_priority_orphan]
         end
 
-        specify '.active.currently_unsponsored should correctly return active unsponsored orphans only' do
```
NikitaAvvakumov: Chaining of scopes is done by AR, no need to test

spec/models/orphan_spec.rb
```diff
@@ -464,23 +434,18 @@
...
-          Orphan.sort_by_eligibility.to_a.each_with_index.map do |orphan, index|
-            expect([active_previously_sponsored_high_priority_orphan, active_previously_sponsored_orphan,
-                    active_unsponsored_high_priority_orphan, active_unsponsored_orphan][index]).to eq orphan
-          end
+          expect(Orphan.sort_by_eligibility).to eq [
+            active_previously_sponsored_high_priority_orphan,
+            active_previously_sponsored_orphan,
+            active_unsponsored_high_priority_orphan,
+            active_unsponsored_orphan
+          ]
```
NikitaAvvakumov: I'd found the original spec difficult to follow.

db/migrate/20150226095809_add_status_enums_to_orphan.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+  end
+
+  def set_sponsorship_status
+    Orphan.all.each do |orphan|
+      status_code = orphan.orphan_sponsorship_status.code
+      orphan.update_column(:sponsorship_status, ( status_code - 1 ))
+    end
+  end
+end
```
NikitaAvvakumov: This migration has to do the important task of preserving data integrity and correctly populating the new enum statuses based on previous associations. I had written a spec for it, but since the test was tied to migration history, it was going to be very brittle. Indeed, after rebasing on latest develop, the spec not only failed but also broke the remainder of the test suite, and I had to take it out. You can see the code [here](https://gist.github.com/6dc9c81bab00c484ca82.git).

app/models/orphan.rb
theB264: It's also a possibility but maybe not recommended to have a task that runs weekly that checks all the `sponsorship_status` against that.

app/models/orphan.rb
theB264: Readability is important.  Usually, if you need to add comments in ruby, something else may be amiss

db/schema.rb
theB264: Resetting the db can be problematic.  Personally I've found that the more you use it, the more likely that there will be migration or other db issues later.  I always prefer `drop`, `create`, `migrate`, `seed`.  I haven't heard that anywhere but I've noticed that.

spec/models/orphan_spec.rb
theB264: I had changed this from that because it was not failing when the same array elements were in a different order.  Not sure why.  So I picked each one out in turn and compared it

db/migrate/20150226095809_add_status_enums_to_orphan.rb
theB264: Is there a way to keep the test maybe by sequentially running each migration until this one has been run, then calling `migration.down` once?  I think this is an important thing to test, it could cripple an entire application if it misbehaved.

app/models/orphan.rb
theB264: :+1:

db/migrate/20150226095809_add_status_enums_to_orphan.rb
NikitaAvvakumov: First, I'm really not sure how to do that programmatically. Even the relatively straight-forward
```ruby
  def migration_has_been_run?(version)
    query = "SELECT version FROM schema_migrations WHERE version = '#{version}'"
    ActiveRecord::Base.connection.execute(query).any?
  end
```
is very brittle and eventually begins to return `false` for reasons I don't understand. And that's when the migration in question is the last one.

Secondly, this and all other migrations will only run once in production since we don't have the luxury of resetting or dropping the db. In development and test, all migrations run before the db is seeded, so the conversion of existing data into a new format is not an issue.

db/schema.rb
NikitaAvvakumov: That's been my experience too. I did go through the whole cycle manually here.

Apropos, given yours and mine observations, I'm very uncertain about advice some people give to delete old migrations.

spec/models/orphan_spec.rb
NikitaAvvakumov: I'll try changing the order and see if it fails as it should.

db/schema.rb
theB264: Yes I would also be skeptical of that advice, unless there's something big I'm totally missing this whole time.  Also I've noticed a gem (`https://github.com/ActsAsParanoid/acts_as_paranoid`) can break migrations as well because it expects columns (`deleted_at`) to be present even if the migration hasn't happened yet.  It hooks this code so deep into AR that it's even called from the model while migrations run.

spec/models/orphan_spec.rb
theB264: See if `expect([1, 2, 3]).to eq [2, 3, 1]` fails or not

NikitaAvvakumov: That fails, as does an array of rearranged orphan records in the test. Not sure why it was passing when you tried it, unless you used `match_array` which ignores order.

app/models/orphan.rb
gmunumel: Second that. :+1:

spec/models/orphan_spec.rb
gmunumel: Did you run the test? Maybe it was failing.

NikitaAvvakumov: Everything is passing. And if it weren't, we'd have to fix it and not comment it out :-)

gmunumel: Yes, I agree. :+1: weird.

db/migrate/20150226095809_add_status_enums_to_orphan.rb
razorcd: I don't see a reason to test migrations unless they are very big to test manually. 
We could write a test to `migrate_down`, populate the DB and then `migrate_up`. And after this goes in production we just remove the test. But this is the same as testing this manually.

Since our production db is small we can clone it and test the migration on our production app locally.

NikitaAvvakumov:
Will merge soon, last chance to review!

gmunumel:
Great work nik. These kind of improvements are always welcome. :+1:

NikitaAvvakumov:
Updated for Orphan CRUD

PR #344, opened 2015-03-06
Fixed - OSRA385 Added filter for name and dates range

gmunumel:
https://osraav.atlassian.net/browse/OSRA-385

### DIFF
--- a/features/aa/aa_features/orphans.feature
+++ b/features/aa/aa_features/orphans.feature
@@ -138,3 +138,21 @@ Feature:
     And I click the "Cancel" button
     Then I should be on the "Show Orphan" page for orphan "Orphan 1"
     And I should not see "Orphan N"
+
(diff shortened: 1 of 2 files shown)


features/aa/aa_features/orphans.feature
```diff
@@ -138,3 +138,21 @@ Feature:
     And I click the "Cancel" button
     Then I should be on the "Show Orphan" page for orphan "Orphan 1"
     And I should not see "Orphan N"
+
+  Scenario: **Bug fix OSRA385** Orphan filter should be done by name
```
NikitaAvvakumov: I'm not sure these thorough tests are necessary since the filter functionality is provided by the AA gem and not coded by us. You could probably dispense with tests altogether here and in sponsors.feature from #341.

PR #347, opened 2015-03-11
OSRA-386 Show no. of sponsorships in sponsor index

NikitaAvvakumov:
screenshot
https://osraav.atlassian.net/browse/OSRA-386

In sponsor index view, in addition to displaying 'Yes'/'No' in the 'Request
fulfilled' column, also show the number of current sponsorships out of the
total requested. e.g. 'Yes (2/2)' or 'No (1/3)'.

(This probably warrants a view spec, but since those are not an option under AA, I don't want to write a heavy-weight feature. Whoever ports this to HQ should test it, though.)

_db schema updates are only due to resetting the database, there are no
changes there pertinent to the PR's functionality._

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -17,12 +17,12 @@ ActiveRecord::Schema.define(version: 20141229224941) do
   enable_extension "plpgsql"
 
   create_table "active_admin_comments", force: :cascade do |t|
-    t.string   "namespace",     limit: 255
+    t.string   "namespace"
     t.text     "body"
-    t.string   "resource_id",   limit: 255, null: false
-    t.string   "resource_type", limit: 255, null: false
+    t.string   "resource_id",   null: false
+    t.string   "resource_type", null: false
     t.integer  "author_id"
-    t.string   "author_type",   limit: 255
+    t.string   "author_type"
     t.datetime "created_at"
     t.datetime "updated_at"
   end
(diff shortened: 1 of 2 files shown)


NikitaAvvakumov:
Client's original request was to show the numbers only if request has not been fulfilled. I thought that it's nice to see them in either case, let's see what they say in review.

db/schema.rb
```diff
@@ -303,17 +303,17 @@
...
 
   add_index "statuses", ["code"], name: "index_statuses_on_code", unique: true, using: :btree
   add_index "statuses", ["name"], name: "index_statuses_on_name", unique: true, using: :btree
 
   create_table "users", force: :cascade do |t|
-    t.string   "email",      limit: 255
+    t.string   "email"
     t.datetime "created_at"
     t.datetime "updated_at"
-    t.string   "user_name",  limit: 255
```
gmunumel: Why all these fields have a limit?

razorcd:
:+1: looks good, works well

db/schema.rb
razorcd: If I remember correctly, these limits were added by default when we upgraded Rails. Looks like the db reset didn't add them again.

NikitaAvvakumov: Exactly. Every time I've reset or dropped and recreated the db since then, the limits got taken out again.

gmunumel: Hmmm. I see. I thought they were a request from the client.

gmunumel:
For me too.

PR #348, opened 2015-03-14
OSRA342 user index pagination new

razorcd:
https://osraav.atlassian.net/browse/OSRA-342

Resubmitted the PR after we had problems with https://github.com/AgileVentures/osra/pull/310

### DIFF
--- a/spec/controllers/hq/users_controller_spec.rb
+++ b/spec/controllers/hq/users_controller_spec.rb
@@ -1,7 +1,7 @@
 require 'rails_helper'
 
 RSpec.describe Hq::UsersController, type: :controller do
-  let(:users) { 3.times.map {build_stubbed(:user)} }
+  let(:users) { build_stubbed_list(:user, 3) }
 
   before :each do
     sign_in instance_double(AdminUser)
--- a/spec/controllers/hq/users_controller_spec.rb
+++ b/spec/controllers/hq/users_controller_spec.rb
@@ -1,5 +1,4 @@
 require 'rails_helper'
-require 'will_paginate/array'
 
 RSpec.describe Hq::UsersController, type: :controller do
   let(:users) { 3.times.map {build_stubbed(:user)} }
--- a/spec/controllers/hq/users_controller_spec.rb
+++ b/spec/controllers/hq/users_controller_spec.rb
@@ -2,6 +2,7 @@ require 'rails_helper'
 require 'will_paginate/array'
 
 RSpec.describe Hq::UsersController, type: :controller do
+  let(:users) { 3.times.map {build_stubbed(:user)} }
 
   before :each do
     sign_in instance_double(AdminUser)
@@ -9,8 +10,10 @@ RSpec.describe Hq::UsersController, type: :controller do
   end
 
   specify '#index' do
-    expect(User).to receive(:paginate).with(page: '2')
+    expect(User).to receive(:paginate).with(page: '2').
+                    and_return( users.paginate(per_page: 2, page: 2) )
     get :index, page: '2'
+    expect(assigns(:users).count).to eq 1
     expect(response).to render_template 'index'
   end
 
@@ -24,7 +27,7 @@ RSpec.describe Hq::UsersController, type: :controller do
     before :each do
       @old_user = FactoryGirl.build_stubbed(:user)
       expect(User).to receive(:find).and_return(@old_user)
-    end    
+    end
 
     specify 'editing renders the edit view' do
       get :edit, id: @old_user.id
@@ -48,7 +51,7 @@ RSpec.describe Hq::UsersController, type: :controller do
   end
 
   describe '#new' do
-    
+
     before :each do
       @new_user = User.new
       allow(User).to receive(:new).and_return(@new_user)
--- a/spec/controllers/hq/users_controller_spec.rb
+++ b/spec/controllers/hq/users_controller_spec.rb
@@ -1,4 +1,5 @@
 require 'rails_helper'
+require 'will_paginate/array'
 
 RSpec.describe Hq::UsersController, type: :controller do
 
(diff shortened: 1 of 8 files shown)


theB264:
Looks good to me :shipit:

gmunumel:
Ready to go :+1:

spec/views/hq/users/index_spec.rb
```diff
@@ -2,51 +2,62 @@
...
+
         expect(rendered).to have_selector('table tr', count: 2)
         expect(rendered).to have_link('Frederick Bloggs', href: hq_user_path(@users[0]))
         expect(rendered).to match 'fred.bloggs@example.com'
-        expect(rendered).to match '5'
       end
     end
 
     context '>1 user' do
+      before :each do
```
NikitaAvvakumov: I'm still not sure why display of multiple users warrants a separate test from display of a single user

spec/views/hq/users/index_spec.rb
```diff
@@ -2,51 +2,62 @@
...
 
     context '1 user' do
       it 'displays a single user' do
-        @users = [ double('User', user_name: 'Frederick Bloggs', 
-          email: 'fred.bloggs@example.com', active_sponsors: (1..5).to_a)]
+        @users = [ FactoryGirl.build_stubbed(:user, user_name: 'Frederick Bloggs',
+          email: 'fred.bloggs@example.com') ].paginate
         render
+
         expect(rendered).to have_selector('table tr', count: 2)
```
NikitaAvvakumov: If there is only one user, why expect 2 `tr`?

spec/views/hq/users/index_spec.rb
```diff
@@ -2,51 +2,62 @@
...
         render
-        expect(rendered).to have_selector('table tr', count: 3)
-        expect(rendered).to have_link('Frederick Bloggs', href: hq_user_path(@users[0]))
-        expect(rendered).to match 'fred.bloggs@example.com'
-        expect(rendered).to match '5'
-        expect(rendered).to have_link('Joseph Soap', href: hq_user_path(@users[1]))
-        expect(rendered).to match 'joe.soap@example.com'
-        expect(rendered).to match '7'
+
+        expect(rendered).to have_selector('table tr', count: 2)
```
NikitaAvvakumov: If I'm reading the setup correctly, 3 users are created, then split into 2 per page, i.e. 2 users on p1 and 1 user on p2. Then the second page is rendered. But it contains two users - ? Also, the test relies on default sorting by id. If we override that later, this test will break for the wrong reasons.

Overall, as my comment above states, I'm not sure why we are testing this at all. Pagination is handled by a gem.

spec/controllers/hq/users_controller_spec.rb
```diff
@@ -1,15 +1,18 @@
 require 'rails_helper'
 
 RSpec.describe Hq::UsersController, type: :controller do
+  let(:users) { 3.times.map {build_stubbed(:user)} }
```
NikitaAvvakumov: Let's use the `_list` methods everywhere

spec/views/hq/users/index_spec.rb
razorcd: because first tr is the table head.
We can use:
`expect(rendered).to have_selector('table tbody tr', count: 1)`

spec/views/hq/users/index_spec.rb
razorcd: because on a single user it doesn't display the pagination buttons. Pagination buttons that looks like we forgot to check!?

spec/views/hq/users/index_spec.rb
gmunumel: Maybe the header?

spec/views/hq/users/index_spec.rb
gmunumel: I like that one :+1:

spec/views/hq/users/index_spec.rb
```diff
@@ -2,51 +2,81 @@
...
-        expect(rendered).to match '5'
-        expect(rendered).to have_link('Joseph Soap', href: hq_user_path(@users[1]))
-        expect(rendered).to match 'joe.soap@example.com'
-        expect(rendered).to match '7'
+
+        expect(rendered).to have_selector('table tbody tr', count: 1)
+        expect(rendered).to have_link('Tom Jones', href: hq_user_path(@users.first))
+        expect(rendered).to match 'tom.jones@example.com'
+        expect(rendered).to_not have_link('Joseph Soap')
+        expect(rendered).to_not match 'joe.soap@example.com'
```
NikitaAvvakumov: Sorry, I still think that this is unnecessary and brittle. If the view renders a single user's info correctly (preceding test), there is no reason to think that it won't render multiples correctly as well. The only thing possibly worth checking is that we included the required fields, and the single-user test accomplishes that. Parsing a collection and displaying each record is handled by Rails and so is not subject to testing here.

Testing that the table body on page 2 only has one row _and_ that it only displays the last user _and_ that the page shows the pagination bar is also redundant, as the three go hand-in-hand.

I also mentioned the issue of brittleness earlier - if we change the sort order of users from id to name or email, the test may break. To make things worse, the first user is created without explicit name or email, so the test may break _sometimes_ depending on `Faker`'s output, which is always a pain to trace.

On a higher level, I believe that this super-tight coupling of tests to objects under test is unnecessarily restrictive. We essentially duplicate the view file in the spec file by iterating over each single field. This kind of brute-force testing - "Let's type the same thing in multiple places so that if we forget to include something in one file, it will be caught by the other" - just feels clumsy and redundant. By that logic, we should duplicate everything again in a feature, to be extra-extra sure. And then have to change 3 files when the client decides not to display user email on the index page.

Our initial aim was to only test logic in view specs, such as conditional displays of elements, and I would really like to stick to that. Which is why I'm making so much noise over this little test :)

razorcd: I see what you mean. This does feel more like an integration test than a simple view unit test.
Also the pagination behavior is tested in the controller, that should be enough.

But don't you think we still need to check that the pagination bar is displayed for multiple users and hidden for 0 or 1 user?

NikitaAvvakumov: To answer questions like that for myself, I ask, "Whose code is making that decision?" If it's ours, test it. If it's from a gem, don't.

razorcd:
pr updated

PR #335, opened 2015-02-20
OSRA-378 Create Rails routes, controller & views for Orphan

christophermlne:
https://osraav.atlassian.net/browse/OSRA-378

Old Status:
—need someone to go over exactly which fields need to go in the edit form, and which model associations, if any, need to be built before an instance of Orphan can be persisted
—do the cucumber specs from aa need to be converted to request specs for hq?

Old Status update:
— just completed a bit of a messy merge which required a rebase.
— need to go through all the code to make sure it still makes sense

Status update:
— ready for code review/merge

PR #353, opened 2015-03-25
OSRA 379, 381, 382, 383 - Orphans CRUD

gmunumel:
Orphans CRUD - OSRA 379, 381, 382, 383. 

Continue from here: https://github.com/AgileVentures/osra/pull/335/files


app/views/hq/orphans/index.html.erb
```diff
@@ -0,0 +1,101 @@
...
+              <%= orphan.date_of_birth %>
+            </td>
+            <td>
+              <%= orphan.gender %>
+            </td>
+            <td>
+              <%= orphan.original_address.province.name %>
+            </td>
+            <td>
+              <%= orphan.original_address.province.name %>
```
gmunumel: How do I get the `Partner` object?

gmunumel:
The error in git hub is related to `Factory Girl`'s random name. Sometimes it fails. Happened to me in my machine, but when I ran the test again it didn't appear. I guess we need to restrict the kind of names `Factory Girl` can use. 

The use of the symbol `O&#39;` is the problem.

razorcd:
Yes, this is one of the problems I wanted to discuss. The view tests are failing for names like 'O'Conner'

app/views/hq/orphans/index.html.erb
razorcd: `orphan.partner` ?  You can leave the eager loading for a different task.

gmunumel: Yes, I see it. Cool. Works great. I thought we need a attribute for `partner` in `orphan` table. But we don't.

spec/views/hq/orphans/show_spec.rb
```diff
@@ -11,7 +11,9 @@
       expect(rendered).to match @orphan.name
```
theB264: Does `@orphan.name` need html escaping in case it contains special characters

NikitaAvvakumov: We have that syntax elsewhere in the suite but it still gives failures on apostrophes

theB264: Are there two different html escaping methods being used?

NikitaAvvakumov: e.g.
```bash
1) hq/partners/_form.html.erb form values
     Failure/Error: assert_select "[value=?]", CGI::escape_html(partner.name)
     Minitest::Assertion:
       Expected at least 1 element matching "[value="O&#39;Connor"]", found 0..
       Expected 0 to be >= 1.
```

NikitaAvvakumov: In fact, it's the `escape_html` that causes the problem. Removing `CGI::escape_html` lets the test pass.

Why are we using Minitest assertions in our RSpec suite?

theB264: No what I mean is the gem or wherever it's coming from using `CGI::escape_html` or is it using some other html escaping method that maybe handles certain characters differently?  I remember at least one test somewhere that wouldn't pass without this.  Or could it be the Minitest assertions working differently than `expect(rendered).to match...`

razorcd: When we added these `CGI::escape_html` I remember that it wouldn't work without it. Names like `O'Conner` would fail withtout CGI::escape.
Now it looks like it's the other way around. Maybe this feature was implemented by default in rails 4.2?

NikitaAvvakumov: My answer to this strangely brittle Minitest
```ruby
assert_select "input#partner_name" do
  assert_select "[value=?]", partner.name # with or without CGI::escape_html
end
```
is Capybara's clean, simple and (in my hands) robust
```ruby
expect(rendered).to have_field('Name', with: partner.name)
```

gmunumel: As a side note. We also need to check `Partners` and `Sponsors` tests. May they have the same problem.

app/controllers/hq/orphans_controller.rb
```diff
@@ -1,6 +1,61 @@
...
+              :orphan_status_id, :priority, :sponsored_by_another_org, :sponsored_minor_siblings_count, 
+              :minor_siblings_count, :sponsored_minor_siblings_count, 
+              original_address: [:id, :city, :province_id, :street, :neighborhood],
+              current_address: [:id, :city, :province_id, :street, :neighborhood])
+  end
+
+  def build_orphan
+    @orphan ||= Orphan.new
+    @orphan.attributes = orphan_params if params[:orphan]
+  end
```
NikitaAvvakumov: I've never been a fan of this method in other HQ controllers. It manages to squeeze two branches into two lines of code, which is a good indication that it does too much. What are the benefits of using it instead of explicit assignments in controller actions? Here in the orphans controller, it is particularly redundant as it is only called from `update`, which means that there is no need for the conditionals: both `@orphan` and `params[:orphan]` will always exist by the time this method gets called. It looks like we can save 4 LOC and eliminate branching by replacing `build_orphan` with `@orphan.attributes = orphan_params` in the `update` action and deleting this method. What do you think?

app/views/hq/orphans/_form.html.erb
```diff
@@ -0,0 +1,157 @@
...
+        <%= f.label :orphan_status_id %>
+        <%= f.collection_select :orphan_status_id,
+          @orphan_statuses, :id, :name, {}, 
+          {class: "form-control"} %>
+      </div>
+      <div class="form-group">
+        <%= f.label :orphan_sponsorship_status_id %>
+        <%= f.collection_select :orphan_sponsorship_status_id,
+          @orphan_sponsorship_statuses, :id, :name, {}, 
+          {class: "form-control", disabled: !f.object.new_record?} %>
```
NikitaAvvakumov: Sponsorship status will not be manually set under any circumstances. I realize that it's a moot point since the form is not used for new records, but there simply is no need for the conditional, the field should be always disabled.

spec/controllers/hq/orphans_controller_spec.rb
```diff
@@ -2,16 +2,47 @@
 require 'will_paginate/array'
 
 RSpec.describe Hq::OrphansController, type: :controller do
-  let(:orphan) {build_stubbed :orphan}
```
NikitaAvvakumov: @gmunumel My comment about using `@` instead of `let` only applied to view specs. Elsewhere, `let` is preferred. Don't worry about changing it back here, it's not a huge deal.

spec/views/hq/orphans/_form_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+
+  specify 'form values' do
+    render
+
+    #fextfields
+    ["name", "father_given_name", "family_name", "mother_name", "date_of_birth",
+     "minor_siblings_count", "sponsored_minor_siblings_count"].each do |field|
+      assert_select "input#orphan_#{field}" do
+        assert_select "[value=?]", CGI::escape_html(@orphan[field].to_s)
+      end
```
NikitaAvvakumov: @gmunumel Are you planning to convert these to Capybara or leave as is?

spec/views/hq/orphans/_form_spec.rb
```diff
@@ -0,0 +1,103 @@
+require 'rails_helper'
+require 'cgi'
+
+RSpec.describe "hq/orphans/_form.html.erb", type: :view do
```
NikitaAvvakumov: Can we have a discussion again about the need for this extensive testing of the form? Rails is handling pre-population of fields with the correct data. Are we double-checking everything just to make sure that we didn't make any typos?

spec/views/hq/orphans/_form_spec.rb
razorcd: We discussed in the last nights meeting that these escaping issues will be fixed in a different task. (all other new and edit views have the same problem).
I noticed there was already a task for this: https://osraav.atlassian.net/browse/OSRA-372

NikitaAvvakumov: :+1:

app/controllers/hq/orphans_controller.rb
razorcd: We don't have the `create` action here and the `build_orphan` is not reused anymore. It can go.

spec/views/hq/orphans/_form_spec.rb
razorcd: We are also checking that select fields are selecting the correct option, text field have correct values ... in the Sponsor view we also check that some fields are not displayed in the form. I think it is easy to make mistakes here so better test it extensively.

gmunumel: @NikitaAvvakumov, i see your point here. But I prefer to test this just in case you are missing something or some field is not fill correctly.

razorcd:
I looked over this PR and made some changes. Once Gabriel will merge them, the PR will be updated.
I looked over everything except the specs for the views.

razorcd:
Some missing fields in the Orphan show view:
created at
updated at
health status
schooling status
goes to school
father occupation
father place of death
father cause of death
guardian (all fields about this)
original address details
current address details
comments

Missing fields in Edit Form:
father occupation
father place of death
father cause of death
father date of death
guardian.. (all 3 fields)
alt. contact number
current address
original address details
contract number
comments

Were these forgotten?

gmunumel:
I added the parameters we display in the view and not all of them.

razorcd:
Just click `Merge` here:
https://github.com/gmunumel/osra/pull/1
and follow instructions

gmunumel:
Guys,  I added the missing fields. However when I tried to test them they always fail. They don't even work in the simplest cases, for example:

```
assert_select "input#orphan_goes_to_school" do
    assert_select "[checked]", @orphan.goes_to_school
end
```
I already check the checkbox in the view and his `id` name. I noticed that fields with validations in`app/models/orphan.rb` like `contact_number` works as expected.

NikitaAvvakumov:
@gmunumel Ensure that the missing attributes are added to the list of permitted params in the controller. If that doesn't fix it, can you share more of your test setup?

Since we've decided to get rid of minitest asserts, you can also do the following:
```ruby
it 'correctly renders checkbox checked or unchecked' do
  orphan = build_stubbed :orphan, :goes_to_school => [true, false].sample

  render partial: ..., locals: {...}

  expect(rendered).to have_field('orphan_goes_to_school',
                                                  checked: @orphan.goes_to_school)
```

gmunumel:
Is this ok to go?

NikitaAvvakumov:
@gmunumel The way we typically handle this is:
- switch to your local develop branch and run `git pull upstream develop` (assuming this repo is the `upstream` remote)
- switch to your orphans_controller branch and run `git rebase develop`
- resolve conflicts as necessary; when you are finished, `git status` will show no changes
- push the rebased orphans_controller branch up to github with `git push -f origin orphans_controller`

If you ran `git pull upstream develop` while in your local orphans_controller branch, you did a merge of upstream/develop into orphans_controller rather than a rebase. This is why you see all the file changes: this workflow necessitates creation of merge commits - i.e. right now you'd do `git commit -am 'Merge develop'` to commit all the changes. If you do that, you can then do an interactive rebase with `git rebase -i HEAD~2` and mark the merge commit for `fixup`, which means to squash into the previous commit. Alternatively, you can reset to your last pre-merge commit and then follow the rebase procedure I described above.

app/views/hq/orphans/_form.html.erb
```diff
@@ -0,0 +1,224 @@
...
+        <%= f.select :priority, 
+          %w(Normal High), {}, class: "form-control" %>
+      </div>
+      <div class="form-group">
+        <%= f.label :created_at %>
+        <%= f.text_field :created_at, class: "form-control date-picker", value: orphan.created_at %>
+      </div>
+      <div class="form-group">
+        <%= f.label :updated_at %>
+        <%= f.text_field :updated_at, class: "form-control date-picker", value: orphan.updated_at %>
```
NikitaAvvakumov: I don't think we want to allow manual updating of `created_at` & `updated_at`. Displaying them is probably useful (although they are not in AA and the client has not complained), but they should not be enabled, nor should they be included in the list of permitted params. It would also be good to give them more user-friendly and OSRA-specific labels, such as "Record imported on" and "Record last updated on".

app/views/hq/orphans/show.html.erb
```diff
@@ -0,0 +1,230 @@
...
+</div>
+<div class="panel panel-default">
+  <div class="panel-heading">
+    <h3 class="panel-title">
+      Additional Details
+    </h3>
+  </div>
+  <div class="panel-body">
+    <dl class="dl-horizontal">
+      <dt>Spnsr Another Organization</dt>
```
NikitaAvvakumov: Is "Spnsr" deliberate? If space is a concern, can we do "Sponsored by Another Org." instead? Or "Sponsored by Third Party"?

spec/views/hq/orphans/_form_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+  specify 'has a "Cancel" button' do
+    render_orphan_form orphan_full
+
+    assert_select 'a[href=?]', hq_orphan_path(orphan_full.id), text: 'Cancel'
+  end
+
+  specify 'has form values' do
+    render_orphan_form orphan_full
+
+    #fextfields
```
NikitaAvvakumov: Please, _t_extfields

spec/views/hq/orphans/show_spec.rb
```diff
@@ -1,17 +1,21 @@
...
   describe 'the orphan exists' do
     before :each do
-      @orphan = build_stubbed :orphan
+      assign(:orphan, orphan)
       render
     end
 
     it 'should show the assigned orphan details' do
-      expect(rendered).to match @orphan.name
+      expect(rendered).to match orphan.name
```
NikitaAvvakumov: I can't decide if this one-line test does enough to be useful at all and whether we might as well get rid of it entirely. Also, the contrast between it and the overly detailed form spec once again makes me wonder about the latter...

spec/views/hq/orphans/_form_spec.rb
NikitaAvvakumov: I agree that it's a good idea to test disabling of inputs (especially if it's conditional), as well as the contents of those inputs that are populated by some kind of custom code. But making sure that Rails correctly fills `:name` with `orphan.name` just seems unnecessary.

spec/views/hq/orphans/_form_spec.rb
gmunumel: What is that means?

spec/views/hq/orphans/show_spec.rb
gmunumel: I can add more fields in the test, but at the end it's just to probe that something is shown.

spec/views/hq/orphans/_form_spec.rb
razorcd: #fextfields   =>    #textfields

gmunumel: hahahaha. lol :+1:

app/views/hq/orphans/_form.html.erb
```diff
@@ -0,0 +1,224 @@
...
+          orphan_sponsorship_statuses, :id, :name, {}, 
+          {class: "form-control", disabled: "true"} %>
+      </div>
+      <div class="form-group">
+        <%= f.label :priority %>
+        <%= f.select :priority, 
+          %w(Normal High), {}, class: "form-control" %>
+      </div>
+      <div class="form-group">
+        <%= f.label :record_imported_on %>
```
NikitaAvvakumov: Here and on L55, this isn't quite how you want to handle the label change. When the attribute for `f.label` and the attribute for the corresponding `f.text_field` match, a user can click on either of them to make the input active and to start entering data. Passing a different attribute to the label than to the input breaks this behaviour.

What you want to do instead is keep the attribute but pass custom text for the label like so:
```erb
<%= f.label, :created_at, 'Record imported on' %>
```

The input is disabled here and so this point is entirely academic, but it's preferable to stick to conventions and best practices nonetheless.

PR #356, opened 2015-03-31
OSRA-397 Sponsorship totals on sponsors index

NikitaAvvakumov:
https://osraav.atlassian.net/browse/OSRA-397

## AA
 - add subtitle to display totals of active & requested sponsorships
 - disable batch actions globally as they are not used by the app
screen shot 2015-03-26 at 1 32 16 pm

## HQ
 - add subtitle to display totals of active & requested sponsorships
 - create view helper methods to calculate the totals (plan to reuse them later
   on a dashboard or elsewhere)
 - view spec for the index page
 - helper spec
screen shot 2015-03-26 at 1 57 10 pm

### DIFF
--- a/app/admin/sponsor.rb
+++ b/app/admin/sponsor.rb
@@ -1,5 +1,7 @@
 ActiveAdmin.register Sponsor do
 
+  actions :all, except: [:destroy]
+
   filter :name, as: :string, label: "Name"
   filter :gender, as: :select, collection: Settings.lookup.gender
   filter :branch, as: :select, collection: proc { Branch.pluck(:name, :id) }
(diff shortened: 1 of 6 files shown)


app/admin/sponsor.rb
```diff
@@ -1,5 +1,19 @@
 ActiveAdmin.register Sponsor do
 
+  actions :all, except: [:destroy]
+
+  total_requested_sponsorships = -> { Sponsor.pluck(:requested_orphan_count).sum }
+  total_active_sponsorships = -> { Sponsorship.all_active.size }
```
razorcd: AA doesn't have access to the helper files?

spec/views/hq/sponsors/index_spec.rb
```diff
@@ -8,6 +8,19 @@
...
+      and_return 3
+    allow(view).to receive(:total_requested_sponsorships).
+      and_return 10
+
+    render
+
+    expect(rendered).
+      to have_text 'Currently sponsoring 3 orphans out of 10 requested.'
+  end
+
```
razorcd: Any reason you are using `allow` instead of `expect`? I favor `expect`.

NikitaAvvakumov: I lean towards the opinion I expressed in the scrum: while nothing would change in practical terms by switching to `expect`, I favour the structure that uses `allow` in the setup phase and `expect` in the verification phase. Using `expect` on L13 & 15 would be testing the implementation, while what I'd rather focus on here is the final outcome - that the right text is rendered.

app/admin/sponsor.rb
NikitaAvvakumov: Not by default, but there is a way to make it happen: http://stackoverflow.com/a/13140275/2623544. I'll push an update shortly. We have other AA helpers that duplicate HQ helper functionality - can either get rid of them too or wait until the whole of AA goes away.

NikitaAvvakumov:
Updated to incorporate @razorcd's suggestion

PR #357, opened 2015-04-01
Osra393 extract date not beyond the first of next month validator

razorcd:
https://osraav.atlassian.net/browse/OSRA-393

Extracted custom validator:  `:date_not_beyond_first_of_next_month`
to `app/models/validators/date_not_beyond_first_of_next_month_validator.rb`

### DIFF
--- a/spec/models/validators/date_not_beyond_first_of_next_month_validator_spec.rb
+++ b/spec/models/validators/date_not_beyond_first_of_next_month_validator_spec.rb
@@ -1,0 +1,34 @@
+require 'rails_helper'
+
+RSpec.describe DateNotBeyondFirstOfNextMonthValidator do
+  let(:test_model) do
+    Class.new do
+      include ActiveModel::Validations
+
+      attr_accessor :date_attr
+
+      validates :date_attr, date_not_beyond_first_of_next_month: true
+    end
+  end
+
+  subject { test_model.new }
+
+  today = Date.current
+  first_of_next_month = today.beginning_of_month.next_month
+  yesterday = today.yesterday
+  second_of_next_month = first_of_next_month + 1.day
+  two_months_ahead = today + 2.months
+
+  [today, first_of_next_month, yesterday].each do |good_date|
+    it { is_expected.to allow_value(good_date).for :date_attr }
+  end
+
+  [second_of_next_month, two_months_ahead].each do |bad_date|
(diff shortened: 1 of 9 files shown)


spec/models/validators/date_not_beyond_first_of_next_month_validator_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'rails_helper'
+include ActiveSupport::Testing::TimeHelpers
```
razorcd: I think it's better to add includes after the `RSpec.describe`. But in this case it makes no difference.

NikitaAvvakumov: I think it's conventional to have all `include`s & `require`s up front

spec/models/validators/date_not_beyond_first_of_next_month_validator_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  it { is_expected.to_not allow_value(second_of_next_month).for :date_attr }
+  it { is_expected.to_not allow_value(two_months_ahead).for :date_attr }
+
+  it 'returns appropriate message when validation fails' do
+    subject.date_attr = two_months_ahead
+    subject.valid?
+    expect(subject.errors[:date_attr]).to eq ['can not be later than the first of next month']
+  end
+
+  it "allows date attribute to be an invalid date format" do
```
NikitaAvvakumov: What does this test tell us, again?

NikitaAvvakumov:
Looks good!

spec/models/validators/date_not_beyond_first_of_next_month_validator_spec.rb
razorcd: If we input some bad date format ("ldjhgldjh") then this validation will pass. It is `:valid_date_presence` validators duty to catch this. Otherwise we'll have too many validation errors.

NikitaAvvakumov: :+1:

NikitaAvvakumov:
Does this need a rebase to get rid of the first 4-5 commits? They look like they've been merged already.

razorcd:
done!

PR #361, opened 2015-04-09
Osra 355 highlight mandatory form fields

NikitaAvvakumov:
https://osraav.atlassian.net/browse/OSRA-355

Monkey patch of Rails' form builder to add `label_for_field` method that applies `class="required_field"` to labels for attributes with presence validations, and then passes the result to the default `label` method.

- two types of validations taken into account: Rails' native `PresenceValidator` and our custom `ValidDatePresenceValidator` (this is stored in a global constant)
- `required_field` CSS class appends ` (required)` and extends `text-danger` for red highlight
- two RSpec custom matchers to test that all required fields, but no optional fields, have been marked
- explicit presence validations added to models where necessary to facilitate marking of required fields
- stylesheet manifest reconfigured as per Bootstrap guidelines
screen shot 2015-04-09 at 7 20 27 pm


NikitaAvvakumov:
@robbiejackson, @PurityControl Since you, respectively, proposed or expressed opinions about the original solution, feedback from both of you would be much appreciated.

PurityControl:
I'm not sure why you would monkey patch FormBuilder and implement a new method `label_for_field` that every one has to remember to use in place of `label` going forward for all forms when you could inherit from FormBuilder and override the `label` method.

NikitaAvvakumov:
@PurityControl And that's why we miss you :)

robbiejackson:
Well I guess my preferred option would still have been to set the Required attribute on the html input control, and then use javascript to set the CSS class of the associated label. That seemed more aligned with HTML5 standards and did allow the browser to validate it.

However, going down the route of the database field validator you'd just have to check if were going to use validations like presence: true with :if or :unless, and whether you wanted this sort of marking on any form fields which didn't map to database fields.

NikitaAvvakumov:
@robbiejackson I hadn't considered the case of presence validators using `if/unless`. In such cases, parsing the logic to determine whether the field ought to be marked might be too much to ask of a view helper.

This actually makes me wonder whether setting inputs to `required` is the best idea - the model can perform such conditional validations, but a browser cannot. If an input got marked as `required` mistakenly in such a case, a user would be prevented from submitting the form even if the data they were trying to send were valid. In general, having two distinct validation stages feels fragile as they are liable to become desynchronized. By contrast, appending a mark to labels is merely a non-binding guideline.

In cases of forms for non-AR/AM objects, I can't think of a solution other than manual markup of labels and/or inputs.

@PurityControl Is your objection to extending the default form builder limited to the fact that we have to use `label_for_field` throughout the form rather than `:builder => OsraFormBuilder` once? What do you think about the fact that with inheritance and overriding, the native `label` method becomes entirely unavailable in forms that use the custom builder, whereas one is free to call `label_for_field` or `label` with the current implementation?

PurityControl:
Given the choice of using inheritance vs monkey patching I would always default to inheritance unless there was a compelling reason to the contrary.

I'm not sure the native label method needs to be lost using inheritance.

To make the new label method the default then setup the following:

```
#app/helpers/application_helper.rb

 module ApplicationHelper
   ActionView::Base.default_form_builder = OsraFormBuilder
 end
```
 
If you want to implement the original behaviour use `:builder => FormBuilder` call directly in the `form_for` call, which I can't really see you using, but is there if you need it.

PurityControl:
Sorry misread your question and didn't pick up on the use of both methods within a single form.
In short can you really see this happening? 
If this was needed I would still look to another solution such as implementing the method in an inherited class.
I don't see anything here that would make me feel a monkey patch is the right direction to go in.

NikitaAvvakumov:
That is, indeed, very unlikely. 

I'm interested in coming up with concrete objections to the monkey patch solution but have yet to find anything that would make me label it as universally undesirable. For the sake of transparency to future OSRA developers (ourselves included), I would steer clear of setting the custom form builder as a global default, and for the same reason do not wish for a monkey patch to override a native method. Specifying a custom form builder for each form should be fairly understandable, but so is use of a custom method whose definition can be looked up with a single click/keystroke in most text editors & IDEs. I suspect that for someone with a bit of Rails experience, an unusual method would ring the bell before they'd notice that a custom form builder is being used.

In terms of remembering to use a custom method vs. a custom form builder, there does not seem to be much difference. One is just as easy to forget as the other, and running a find-and-replace on a file is not much more work than specifying a form builder to be used.

PurityControl:
At the end of the day you need to go for the solution that feels most natural for the team.

However, I think looking for concrete objections to the monkey patch is a red herring, in so far as an absence of objections doesn’t make it necessarily the right thing to do.

My main objection is cost of change.

Most people agree that, for objection orientated design, abstraction of behaviour should be through an interface. Only when there is good reason to prefer inheritance should you use inheritance.

in this case inheritance would be be a better choice as the new OsraFormBuilder is a type of FormBuilder.

Taking this further I would only prefer a monkey patch over inheritance if there is a good reason to do so. Monkey Patching almost always seems harmless at the time and the technical debt only seen when its too late to avoid paying it back. There is a reason one of rspec’s goals 2 -> 3 -> 4 is to move to zero monkey patching.

In this case I can’t see any benefit to using monkey patching over other more conventional means and for that reason alone I believe it should be ruled out.

I totally understand your objection to obscurity and not wanting to override the default FormBuilder in settings. I think this also applies to the monkey patch though. I think because of this the new method should be in a class that extends the FormBuilder. You can then declare this CustomFormBuilder in the form_for, making everything explicit and clear in purpose / intent.

I think the argument for being able to find out where a custom method is defined via any one experienced with a text editor is perfectly valid but counter argues the point of wanting transparency. We extend ActiveRecord::Base when we want to implement a type of model with new functionality. Similarly ActionController::Base when implementing controllers. This pattern of extending functionality in this way cascades all through our rails apps and for good reason. Why does this feel so wrong for the FormBuilder and the use of a monkey patch feel so right?

NikitaAvvakumov:
I have no feelings of right or wrong regarding monkey patch vs. inheritance, but am trying to find objective arguments to support either choice. Your point above is very well taken.

razorcd:
My opinion is that creating a custom FormBuilder class just to add one simple functionality would be too much here. For some complex logic it would make sense but since we are talking about the View I wound't worry about it. 

Though for a Model related class I would prefer to create a custom class using inheritance, DI, etc. just because it will scale in complexity a lot better.

So I would stick with monkey patching here. Probably with more experience I will see things differently.

Also just extending the class it is more newbie friendly and since we want developers to join us in the future, we should consider this.

gmunumel:
I'm a bit new in these topics, however I get the whole idea. I've been reading on internet and trying to get the good and bad of using monkey-patching or an inheritance approach. General discussing are focus in how risky could be using monkey-patching when you change a given behaviour and that you need to remember that you are doing it so. While using inheritance is just the more observable way.  

My humble opinion is to use the approach that give us more advantages at the end. Of course, always thinking about others programmers would be dealing with this in the future. 

At the end if we use monkey-patching make a mention in a document or a comment in the code.

NikitaAvvakumov:
Everyone, thank you for your feedback. After researching and thinking about the main issue some more, I made the following change to this PR: the new `label_for_field` is now defined in a module, which is included in `ActionView::Helpers::FormBuilder` on a per-form basis and not globally as before. This was inspired by [this article](http://www.justinweiss.com/blog/2015/01/20/3-ways-to-monkey-patch-without-making-a-mess/).

Once I made the above change, the monkey patch solution became essentially the same as using inheritance. Given that, I've just submitted PR #363, in which a custom form builder is subclassed from `ActionView::Helpers::FormBuilder`.

Regardless of the approach taken, I still strongly prefer not to make the change a global default but to make it explicit in each form; and not to override any existing methods but rather to add a custom one.

gmunumel:
Totally agree with the desicion. :+1: 

But your idea got me thinking. How can you make this change globally?

NikitaAvvakumov:
Superseded by #363

gmunumel:
Yep. I saw that :+1: 
 :)

PR #362, opened 2015-04-13
Osra372 view specs intermittent failures

razorcd:
https://osraav.atlassian.net/browse/OSRA-372

Changed `minitest` assertions to `rspec+capybara` for all `forms` in the view specs.


spec/views/hq/sponsors/_form_spec.rb
```diff
@@ -24,129 +23,95 @@ def render_sponsor_form current_sponsor
...
-    assert_select 'form'
+    expect(rendered).to have_selector("form")
   end
 
   describe 'has a "Cancel" button' do
     specify 'using an existing Sponsor record' do
       render_sponsor_form sponsor_full
 
-      assert_select 'a[href=?]', hq_sponsor_path(sponsor_full.id), text: 'Cancel'
+      expect(rendered).to have_link("Cancel", hq_sponsor_path(sponsor_full.id))
```
NikitaAvvakumov: Let's use this opportunity to get rid of the hard-coded `id`.

spec/views/hq/sponsors/_form_spec.rb
```diff
@@ -24,129 +23,95 @@ def render_sponsor_form current_sponsor
...
     end
   end
 
   describe 'has form values' do
     specify 'using an existing Sponsor record' do
-      allow(User).to receive(:pluck).and_return([user.user_name, user.id])
+      allow(User).to receive(:pluck).and_return([[user.user_name, user.id]])
       render_sponsor_form sponsor_full
 
       #fextfields
```
NikitaAvvakumov: Can we finally correct this? :)

spec/views/hq/sponsors/_form_spec.rb
```diff
@@ -24,129 +23,95 @@ def render_sponsor_form current_sponsor
...
-          assert_select "option",
-            html: CGI::escape_html(sponsor_full.branch.name)
-        else
-          assert_select "option", value: false
-        end
+      if sponsor_full.branch
+        expect(rendered).to have_selector("select[id='sponsor_branch_id'][disabled] option[selected]", text: sponsor_full.branch.name)
+      else
+        expect(rendered).to have_selector("select[id='sponsor_branch_id'][disabled] option", text: "")
+        expect(rendered).to_not have_selector("select[id='sponsor_branch_id'][disabled] option[selected]")
```
NikitaAvvakumov: It seems that this ends up testing two separate things - the fact that fields are correctly pre-filled with object's attribute values (the form's responsibility) and that organization and branch cannot co-exist for the same sponsor (the model's responsibility). Shouldn't we be able to rely on the model to do its job and only test that if a branch is specified, the relevant field is pre-filled, and likewise for organization?

Also, since both the organization and the branch selectors should be disabled regardless of which affiliation a sponsor has, I think this ought to be tested outside of any `if`.

spec/views/hq/sponsors/_form_spec.rb
```diff
@@ -24,129 +23,95 @@ def render_sponsor_form current_sponsor
...
 
-      assert_select "select#sponsor_agent_id" do
-        assert_select "option", value: sponsor_full.agent_id
-        if sponsor_full.agent
-          assert_select "option",
-            html: CGI::escape_html(sponsor_full.agent.user_name)
-        else
-          assert_select "option", value: false
-        end
+      if sponsor_full.agent
```
NikitaAvvakumov: Agent assignment is now mandatory for all sponsors, no need for the conditional.

spec/views/hq/sponsors/_form_spec.rb
gmunumel: Why is an array within an array?

spec/views/hq/sponsors/_form_spec.rb
razorcd: Why do you say that this is testing that organization and branch can not coexist? It is only testing the views responsability, that it is showing the field correctly.

NikitaAvvakumov: Sorry, I'd misread the code.
My second point remains valid, though.

spec/views/hq/sponsors/_form_spec.rb
razorcd: @gmunumel : `.plunk` returns an array of arrays if you specify more than one field

razorcd:
I moved `model_helpers_spec.rb` from `spec/support` to a new folder: `spec/support_spec` to avoid loading it when we run `rspec` on any other spec file.

Not sure if `spec/support_spec/model_helpers_spec.rb` is the best place for this.

NikitaAvvakumov:
@razorcd I'm not sure that model_helpers_spec or support_spec describe the role of these tests as accurately as possible. What do you think about my approach in #363 - spec/custom_matchers/optional_form_fields_matcher_spec.rb?

razorcd:
Yes, it's better but I don't want the folder that contains the Rspec custom matchr specs to be confused with folders that have App specs. 
How about we use: `spec/RSpec_custom_matchers/...` ?

NikitaAvvakumov:
Sure, just keep it all lower-case to be safe

razorcd:
pr updated

gmunumel:
Great work guys. :+1: 

I like the idea :D

PR #363, opened 2015-04-16
Osra 355 v2 highlight mandatory form fields

NikitaAvvakumov:
Alternative implementation to #361

https://osraav.atlassian.net/browse/OSRA-355

Subclass custom form builder from `ActionView::Helpers::FormBuilder` and define `label_for_field` method within it. The method checks whether the field's attribute has a presence validation and, if so, appends a `required_field` CSS class to the options before passing things along to the native FormBuilder's `label` method. CSS then appends " (required)" to the label.

 - both Rails' native presence validation and OSRA's custom `ValidDatePresence` validation used to determine required attributes
 - custom RSpec matchers to test that required fields are labeled and that optional fields are not
 - explicit presence validations added to several attributes to facilitate labeling of form fields
 - stylesheet reconfigured as per Bootstrap guidelines

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -1,5 +1,3 @@
- @import "bootstrap-sprockets";
- @import "bootstrap";
 /*
  * This is a manifest file that'll be compiled into application.css, which will include all the files
  * listed below.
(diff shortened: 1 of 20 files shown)


gmunumel:
Monkey patching doesn't like so bad either. At this point I don't have a preference with one approach or the other. They look similar.

NikitaAvvakumov:
Updated for HQ Orphan CRUD

PR #365, opened 2015-04-22
osra405 Hq admin_user authentication

razorcd:
https://osraav.atlassian.net/browse/OSRA-405

- added Rails Login/Logout functionality using devise gem
- changed devise messages from using `Sign in`/`Sign out` to use `Log in`/`Log out`
- removed AA login functionality (had to)
- replaced HQ cucumber features with rspec features and moved feature methods in modules
- added login/logout route specs
- small cosmetic css changes


app/models/admin_user.rb
```diff
@@ -1,6 +1,6 @@
 class AdminUser < ActiveRecord::Base
   # Include default devise modules. Others available are:
   # :confirmable, :lockable, :timeoutable and :omniauthable
-  devise :database_authenticatable, 
-         :recoverable, :rememberable, :trackable, :validatable
+  devise :database_authenticatable, :rememberable
```
gmunumel: `:memorable` instead of `:rememberable`

app/views/layouts/_header.html.erb
```diff
@@ -2,7 +2,7 @@
   <div class="container-fluid">
     <!-- Brand and toggle get grouped for better mobile display -->
     <div class="navbar-header">
-      <a class="navbar-brand" href="#">OSRA Support System</a>
+      <a class="navbar-brand" href="/">OSRA Support System</a>
```
gmunumel: Is this causing a real impact? With `/` is the page loaded again?

app/models/admin_user.rb
razorcd: It is `:rememberable`, from devise. It is for the "remember me" checkbox at login form and saving the token. See more here: https://github.com/plataformatec/devise

app/views/layouts/_header.html.erb
razorcd: It points to the root which in our case loads the AA OSRA dashboard.
I thing it's more of a convention to link the header logo to the homepage. Every public website does this.

razorcd:
loginhqform

loginhqform2

loginhqform3

loginhqform4

razorcd:
By default it will load the hq login form. (AA login form/routes removed) But once logged in, it will redirect to AA dashboard.

razorcd:
Notice: because we are not showing the left navbar, the `success` flash message: `Logged out successfully.` is a little off. 
An easy fix would be to just move the flash message just between the header and navbar and remove the border and margin, so it looks more like in AA:

aa

gmunumel:
Looks great :+1:

razorcd:
Hq::Devise and Admin::Devise now run in parallel.   ActiveAdmin functionality is not affected.

app/models/admin_user.rb
gmunumel: Yes. Sorry for that silly question :(

gmunumel:
:shipit:

config/initializers/devise.rb
```diff
@@ -224,7 +224,7 @@
   # config.navigational_formats = ['*/*', :html]
 
   # The default HTTP method used to sign out a resource. Default is :delete.
-  config.sign_out_via = :delete
+  config.sign_out_via = :get
```
NikitaAvvakumov: Why this change? Ideally, `GET` requests should be idempotent, while `DELETE` seems very appropriate here.

config/locales/devise.en.yml
```diff
@@ -41,15 +41,15 @@ en:
...
+      signed_out: "Logged out successfully."
     unlocks:
       send_instructions: "You will receive an email with instructions about how to unlock your account in a few minutes."
       send_paranoid_instructions: "If your account exists, you will receive an email with instructions about how to unlock it in a few minutes."
-      unlocked: "Your account has been unlocked successfully. Please sign in to continue."
+      unlocked: "Your account has been unlocked successfully. Please log in to continue."
   errors:
     messages:
-      already_confirmed: "was already confirmed, please try signing in"
+      already_confirmed: "was already confirmed, please try loggin in"
```
NikitaAvvakumov: '.. logging in'

spec/features/hq/admin_user_authentication_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  scenario 'Restricted pages should be accessible to logged in users' do
+    i_sign_in_as_admin
+    and_i_should_be_on_root_page
+  end
+
+  scenario 'Restricted pages should not be accessible to logged out users' do
+    i_sign_in_as_admin
+    i_sign_out
+    and_i_should_be_on_new_admin_user_session_page
```
NikitaAvvakumov: Minor point: especially in a feature test, it's probably a good idea to use more user-oriented terms. Here, `and_i_should_be_on_sign_in_page` rather than `new_admin_user_session_page`.

spec/features/hq/user_is_creating_multiple_sponsors_spec.rb
```diff
@@ -16,18 +16,11 @@
   end
 end
 
+
```
NikitaAvvakumov: ?

spec/features/hq/user_is_creating_multiple_sponsors_spec.rb
```diff
@@ -16,18 +16,11 @@
...
 def an_agent_exists
   FactoryGirl.create :agent
 end
 
-def i_sign_in_as_admin
```
NikitaAvvakumov: How come this isn't needed anymore?

razorcd: just moved to `spec/features/hq/admin_user_authentication_spec.rb` together with all auth methods

config/initializers/devise.rb
razorcd: because AA is using the `GET` for the `SignOut` link  from header.
But I prefer DELETE also

spec/features/hq/user_is_creating_multiple_sponsors_spec.rb
NikitaAvvakumov: I'm still confused: How are these method definitions shared between files? Can the 'user_is_creating...' feature run on its own?

razorcd: good point. i think these methods should go in a helper module

razorcd:
UPDATE: rspec Feature tests have helper modules for the shared methods.

razorcd:
The only inconvenient for this authorization in rails is that you need to log in in `/admin ` to access the active admin pages and the log in in `/hq` to access the hq pages.
AA and Rails should run in parallel so I think we should keep this behavior.

config/routes.rb
```diff
@@ -1,6 +1,9 @@
 Osra::Application.routes.draw do
 
+  devise_for :admin_users, path: "hq", as: "hq", path_names: { sign_in: 'login', sign_out: 'logout' },
+                           :controllers => { :sessions => "hq/devise/sessions" }
   namespace :hq do
+    root to: "sponsors#index"                 #should be changed to "dashboard" when implemented"
```
theB264: Instead of a comment here, I'd recommend a spec marked pending.  When it passes, it will notify you to fix it.  You can put this comment inside the routing spec for the dashboard view.

robbiejackson:
Christian just checking up with you regarding the folder structure. For the controllers you've got devise under app/controllers/hq as app/controllers/hq/devise 
but for the views you've app/views/devise with hq in the middle between views and devise. Is this what you think is best?

spec/routing/hq/dashboard_routing_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+RSpec.describe "Hq::Dashboard redirecting", type: :request, pending: true do
+  before :each do
+    sign_in instance_double(AdminUser)
+  end
+
+  it "root redirects to dashboard" do
+    get '/hq'
+    expect(response).to redirect_to hq_dashboard_path
+  end
+end
```
razorcd: @theB264 Yes, using pending specs looks more professional. Thanks.
I created this spec but when I run it, `sign_in instance_double(AdminUser)` is giving me an error:
`Could not find a valid mapping for #<RSpec::Mocks::InstanceVerifyingDouble:0x307cee0 @name="AdminUser (instance)">`
I did research to fix it but no luck. Do you have any idea how to make this work?

There is some info about this on Devise docs but not very helpful:
`@request.env["devise.mapping"] = Devise.mappings[:user]`

razorcd:
@robbiejackson yes, you are right. I moved `app/views/devise` to `app/views/hq/devise`

Thanks. Good find!

spec/routing/hq/dashboard_routing_spec.rb
theB264: What I meant is instead of a comment in the routes file which can be forgotten or overlooked, you could put something like this in, and it will force you to edit code and read the comment once the route exists.
``` ruby
  RSpec.describe 'Hq::Dashboard redirecting', type: :routing do
    specify 'disposable spec', pending: 'be sure to route the default "/hq" route here once available' do
      expect(get: '/hq/dashboard').to be_routable
    end
  end
```

razorcd: I see. This is not testing the `redirect` (/hq to /hq/dashboard) but it will remind the developer when it is working on dashboard.

PR #368, opened 2015-04-27
fixed migration for missing_method on orphan_status and orphan_sponsorsh...

razorcd:
Problem with ENUM migration:
- because OrphanStatus and OrphanSponsorshipStatus models were deleted the migration can't access them.

! still WIP, will fix later

### DIFF
--- a/db/migrate/20150226095809_add_status_enums_to_orphan.rb
+++ b/db/migrate/20150226095809_add_status_enums_to_orphan.rb
@@ -15,16 +15,24 @@ class AddStatusEnumsToOrphan < ActiveRecord::Migration
   private
 
   def set_status
+    connection = ActiveRecord::Base.connection
+    orphan_statuses = connection.execute("SELECT * FROM orphan_statuses")
+
     Orphan.all.each do |orphan|
-      status_code = orphan.orphan_status.code
+      orphan_status = orphan_statuses.select {|os| os["id"] == orphan.orphan_status_id.to_s}.first
+      status_code = orphan_status["code"].to_i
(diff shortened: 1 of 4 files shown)


db/migrate/20150226095809_add_status_enums_to_orphan.rb
```diff
@@ -15,16 +15,24 @@ def down
...
+
     Orphan.all.each do |orphan|
-      status_code = orphan.orphan_sponsorship_status.code
-      orphan.update_column(:sponsorship_status, ( status_code - 1 ))
+      orphan_sponsorship_status = orphan_sponsorship_statuses.select {|oss| oss["id"] == orphan.orphan_sponsorship_status_id.to_s}.first
+      sponsorship_status_code = orphan_sponsorship_status["code"].to_i
+      orphan.update_column(:sponsorship_status, ( sponsorship_status_code - 1 ))
     end
   end
 end
```
razorcd: Only this file is important for the FIX, but the rest will not hurt.

NOTICE that `migrate:down` will not populate the `orphan_status_id` and `orphan_sponsorship_status_id` columns back. I don't think that is possible because we don't have the OrphanStatus and OrphanSponsorshipStatus tables and seeds. 
Backup the db before running the migrations. (https://www.digitalocean.com/community/tutorials/how-to-backup-postgresql-databases-on-an-ubuntu-vps)

gmunumel:
Why were OrphanStatus and OrphanSponsorshipStatus models deleted?

razorcd:
Yes, we were replacing them with enums (see Orphan model).

PR #373, opened 2015-05-16
OSRA408 AdminUser #new and #create done, plus small fixes to #index

robbiejackson:


spec/views/hq/admin_users/new.html.erb_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    render
+    expect(view).to render_template(:new)
+  end
+
+  it 'cancel button goes to index page' do
+    render
+    expect(rendered).to have_link('Cancel', href: hq_admin_users_path)
+  end
+
+end
```
razorcd: Should we also test that the view has a form, has all the 3 input fields and has a submit button?

spec/features/hq/create_admin_users.rb
```diff
@@ -0,0 +1,17 @@
...
+
+RSpec.feature 'Creating Admin Users', :type => :feature do
+
+  scenario 'After creation new Admin Users should be able to log in' do
+    i_sign_in_as_admin
+    visit new_hq_admin_user_path
+    fill_in 'Email', with: 'newadmin.user@example.com'
+    fill_in 'Password', with: 'Password'
+    fill_in 'Password confirmation', with: 'Password'
+    click_button 'Create Admin user'
```
razorcd: should we have a `i_create_an_admin_user_with_credentials(email, pass)` for these 4 lines too?

spec/features/hq/create_admin_users.rb
```diff
@@ -0,0 +1,17 @@
...
+    i_sign_out
+    i_sign_in_with_credentials('newadmin.user@example.com', 'Password')
+    expect(page.find('.flashes')).to have_text "Logged in successfully"
```
razorcd: i think `and_i_should_see "Logged in successfully"` will do

razorcd:
please add an image with the form

spec/views/hq/admin_users/new.html.erb_spec.rb
robbiejackson: Yes I'll add those.

spec/features/hq/create_admin_users.rb
robbiejackson: My preference would be to leave these 4 lines here I think. 
What we're testing is that our app is integrated with Devise correctly so that the back-end processing of this form works as expected, so it's particularly related to the completion of the fields on this form. Hence I think having the detail is appropriate.
Also, I don't think that a method such as `i_create_an_admin_user_with_credentials(email, pass)` would be reused in a different context. (You'd just use a factory and sign in instead).
What do you reckon?

spec/features/hq/create_admin_users.rb
robbiejackson: Yes, agreed, I'll change that.

robbiejackson:
new-admin-user

spec/controllers/hq/admin_users_controller_spec.rb
```diff
@@ -3,6 +3,7 @@
 
 RSpec.describe Hq::AdminUsersController, type: :controller do
   let(:admin_users) { build_stubbed_list :admin_user, 2 }
+  let(:admin_user) { admin_users.first }
```
razorcd: Suggestion: By `using build_stubbed :admin_user` instead of `admin_users.first` we are not creating the second unnecessary `admin_user` from `build_stubbed_list`. Also decoupling things as much as possible is a good practice.

razorcd:
in the header we have the email link for the logged in user that points to `admin_user#show`. We should just show the email as plain text. 

Other than these: :shipit:

spec/controllers/hq/admin_users_controller_spec.rb
robbiejackson: Yes I wondered about which to use, and in the end went for the one that was slightly more efficient. But reconsidering it I think you're right about it being cleaner to use a separate build_stubbed, so I've changed that.

robbiejackson:
Good spot about the header - I hadn't noticed that. It's now just text.

gmunumel:
Why are there three PR with the same functionality?

razorcd:
looks good :shipit:

robbiejackson:
Just noticed that the view #new spec was wrongly named - have fixed this.

PR #374, opened 2015-05-18
Osra409 adminuser edit and update (plus minor fix for index)

robbiejackson:
adminuser-edit


spec/features/hq/create_admin_users_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+    fill_in 'Email', with: 'newadmin.user@example.com'
+    fill_in 'Password', with: 'Password'
+    fill_in 'Password confirmation', with: 'Password'
+    click_button 'Create Admin User'
+    i_sign_out
+    i_sign_in_with_credentials('newadmin.user@example.com', 'Password')
+    and_i_should_see "Logged in successfully"
+  end
+
+end
```
razorcd: We have both `spec/features/hq/create_admin_users.rb` and `spec/features/hq/create_admin_users_spec.rb` ?

Also we could use integration tests for edit/update too.

robbiejackson: I thought I'd got rid of that file!
I was trying to do an integration test for that, but was having problems with Factory stubs accessing the database. Let's have a chat about it in the scrum.

robbiejackson:
Re labels for edit being New Password, the problem is that because of how Nik has coded the solution for setting the Required indicator on labels, this will show up as the New Password being required. Which it isn't. I can't think of any sensible way of changing the functionality so that the Required doesn't appear in the label for this case. If we could do that, then New Password would be better, but if we can't, I think it's better just sticking with Password, and the placeholder text there makes it appear that it's already present. 
Also I was wondering about getting the user to reenter his password for some of these operations - particularly the delete. Let's discuss in the scrum.

app/views/hq/admin_users/edit.html.erb
```diff
@@ -0,0 +1,22 @@
+<h1> Edit Admin User </h1>
+
+<%= form_for [:hq, @admin_user], :builder => OsraFormBuilder do |f| %>
+  <%= render 'shared/errors', object: f.object %>
+  <div class="form-group">
+    <%= f.label_for_field :email %>
+    <%= f.text_field :email, class: "form-control" %>
+  </div>
+  <div class="form-group">
+    <%= f.label_for_field :password %>
```
razorcd: so if you do `f.label_for_field :password, "New Password"` it will not work?

robbiejackson: That will change the text of the label to New Password ok. But the red "(required)" will still appear after it, because that's done in CSS styling based on the associated field being mandatory in the model.

razorcd: Ah, I see. Then `f.label :password, "New Password"` should remove the required.

robbiejackson: I'm afraid not.

gmunumel:
Totally agree with this: Also I was wondering about getting the user to reenter his password for some of these operations - particularly the delete.

We should double check if the user is who it says to be. Also many web sites do this doble check to change passwords or do another critical operations.

robbiejackson:
OK, this has now got fixed all the issues from previous code reviews. 
I haven't included the aspect of forcing the admin user to re-enter their password when they're executing certain actions - this is dependent upon what Christian agrees with the clients.
The code here includes the code changes for OSRA408 because the branch was taken off that, but a summary of the files different between OSRA408 and this one OSRA409 is below:
M	app/controllers/hq/admin_users_controller.rb
A	app/views/hq/admin_users/edit.html.erb
M	app/views/hq/admin_users/index.html.erb
M	spec/controllers/hq/admin_users_controller_spec.rb
A	spec/features/hq/edit_admin_users_spec.rb
A	spec/views/hq/admin_users/edit_spec.rb
M	spec/views/hq/admin_users/index_spec.rb

gmunumel:
What the client decided about this functionality?

razorcd:
@gmunumel : we'll have a meeting with the client once we finish rails part. I expect in max 2 weeks!?
@robbiejackson: can you please rebase this so commits: `AdminUser #new and #create done, plus small fixes to #index` will dissapear from this PR.
Thanks

robbiejackson:
This has now been rebased after the merge of OSRA408 (new/create).
New screenshot too:
screenshot from 2015-05-25 21 20 18

razorcd:
:+1:

PR #375, opened 2015-05-18
Osra410 adminuser delete

robbiejackson:


spec/views/hq/admin_users/new.html.erb_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    render
+    expect(view).to render_template(:new)
+    expect(rendered).to have_selector("form")
+    ["email", "password", "password_confirmation"].each do |field|
+      expect(rendered).to have_selector("input[id='admin_user_#{field}']")
+    end
+    expect(rendered).to have_selector("input[type='submit'][value='Create Admin User']")
+  end
+
+  it 'has a cancel button going to the index page' do
```
gmunumel: Why are we testing only a cancel button?

robbiejackson: That's the only other button there is. 
The submit button is tested in the previous test, where it checks that all the standard form elements (form tag, fields and submit button) are present.

razorcd:
please rebase this one too

robbiejackson:
OSRA 415 AdminUser delete now rebased.

PR #379, opened 2015-05-21
Osra 424 Show partner info on orphan pages

ZmagoD:
orphan_partner

### DIFF
--- a/app/views/hq/orphans/show.html.erb
+++ b/app/views/hq/orphans/show.html.erb
@@ -210,7 +210,7 @@
   <div class="panel-body">
     <dl class="dl-horizontal">
       <dt>Partner</dt>
-      <dd><%= @orphan.orphan_list.partner.name %></dd>
+      <dd><%= @orphan.partner.name %></dd>
     </dl>
     <dl class="dl-horizontal">
       <dt>Sponsored by Another Org.</dt>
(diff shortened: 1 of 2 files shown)


gmunumel:
I don't know if this change apply for a test. But if you want to add it please take a look at this: https://github.com/AgileVentures/osra/blob/develop/spec/views/hq/orphans/_form_spec.rb

razorcd:
please remove all extra commits here.
You should only have:
`adding partner to show.html.erb of orphan`
`added test for partner field`

razorcd:
other than that it's good

app/views/hq/orphans/show.html.erb
```diff
@@ -209,6 +209,10 @@
   </div>
   <div class="panel-body">
     <dl class="dl-horizontal">
+      <dt>Partner</dt>
+      <dd><%= @orphan.orphan_list.partner.name %></dd>
```
razorcd: please use `@orphan.partner.name` instead because we have the relation set directly in the model with `has_one :partner, through: :orphan_list, autosave: false`

spec/views/hq/orphans/show_spec.rb
```diff
@@ -12,6 +12,11 @@
     it 'should have an Edit Orphan button' do
       expect(rendered).to have_link('Edit Orphan', edit_hq_orphan_path(orphan.id))
     end
+    
+    
+    it 'should have partner field' do
+      expect(rendered).to have_selector("dd", text: orphan.orphan_list.partner.name)
```
razorcd: same here `orphan.partner.name` 
and please remove empty line 16

PR #380, opened 2015-05-21
Dashboard section of osra

achock:


config/routes.rb
```diff
@@ -1,9 +1,14 @@
 Osra::Application.routes.draw do
 
+  namespace :hq do
+  get 'dashboard/index'
+  end
```
razorcd: why is this needed since we have line 11?

spec/views/hq/dashboard/index.html.erb_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'rails_helper'
+
+RSpec.describe "hq/dashboard/index.html.erb", type: :view do
```
gmunumel: Be careful with the indentation. Try to use a 2 or 4 space indent.

app/controllers/hq_controller.rb
```diff
@@ -1,7 +1,7 @@
 class HqController < ApplicationController
   NAVIGATION_BUTTONS= [
     { text: 'OSRA', href: 'root_path', path_regex: /^\/$/, glyph: 'glyphicon-home' },
-    { text: 'Dashboard', href: '"/404.html"', path_regex: /^\/hq(\/)?$/, glyph: 'glyphicon-dashboard' },
+    { text: 'Dashboard', href: '"dashboard"', path_regex: /^\/hq\/dashboard/, glyph: 'glyphicon-dashboard' },
```
gmunumel: You can change `"dashboard"` with  `hq_dashboard_path` since you created the route for it. If you are not sure about the route please take a look at `rake routes`.

app/assets/javascripts/hq/dashboard.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
```
gmunumel: I don't see the point of adding a `.coffee` file that we are not using it at the moment. My opinion is to added only when  we *really* need it.

app/assets/stylesheets/hq/dashboard.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the hq/dashboard controller here.
```
gmunumel: Same opinion as the above.

spec/controllers/hq/dashboard_controller_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'rails_helper'
+
+RSpec.describe Hq::DashboardController, type: :controller do
+
+  describe "GET #index view" do
+  	before(:each) do
+    	sign_in instance_double(AdminUser)
```
gmunumel: Please be careful with the indentation.

app/helpers/hq/dashboard_helper.rb
```diff
@@ -0,0 +1,2 @@
+module Hq::DashboardHelper
+end
```
gmunumel: Is this *really* need it?

achock: Most of these are generated by the generator. So this is why they are there.

gmunumel: Ok. So I guess we can get rid of them and add them only in case we need them.

PR #381, opened 2015-05-26
Osra418 sort by table header for orphans

robbiejackson:
This functionality for sorting by headers follows the design proposed in http://railscasts.com/episodes/228-sortable-table-columns and the first part of http://railscasts.com/episodes/240-search-sort-paginate-with-ajax.
There will be commonality between this sorting for Orphans and sorting other tables, so please comment particularly on these aspects so that we can agree on their design. They are:
1. The use of the helper function sortable_column
2. The use of the private methods #sort_column and #sort_direction in the controllers (to prevent SQL injection attacks)
3. The CSS class we use to indicate the current sorting column and direction (asc/desc) 
(eg we might instead want to have 2 classes one for the column and one for the direction, as shown in the railscast).
(I've just created temporarily a little CSS in /assets/stylesheets/hq.scss to make the background of the column header green if it's an ascending sort, pink if descending).

### DIFF
--- a/app/controllers/hq/orphans_controller.rb
+++ b/app/controllers/hq/orphans_controller.rb
@@ -6,15 +6,7 @@ class Hq::OrphansController < HqController
   ADDRESS_DETAILS = [:id, :city, :province_id, :street, :neighborhood, :details]
 
   def index
-    if params[:sort_by] == 'original_address_province_name'
-      @orphans = Orphan.includes(:original_address).order("provinces.name " + sort_direction).paginate(:page => params[:page])
-    elsif params[:sort_by] == 'orphan_list_partner_name'
-      @orphans = Orphan.includes(:orphan_list).includes(:partner).order("partners.name " + sort_direction).paginate(:page => params[:page])
-    elsif sort_column
-      @orphans = Orphan.order(sort_column + ' ' + sort_direction).paginate(:page => params[:page])
-    else
-      @orphans = Orphan.paginate(:page => params[:page])
-    end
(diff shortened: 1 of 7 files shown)


app/controllers/hq/orphans_controller.rb
```diff
@@ -1,9 +1,20 @@
...
+    if params[:sort_by] == 'original_address_province_name'
+      @orphans = Orphan.includes(:original_address).order("provinces.name " + sort_direction).paginate(:page => params[:page])
+    elsif params[:sort_by] == 'orphan_list_partner_name'
+      @orphans = Orphan.includes(:orphan_list).includes(:partner).order("partners.name " + sort_direction).paginate(:page => params[:page])
+    elsif sort_column
+      @orphans = Orphan.order(sort_column + ' ' + sort_direction).paginate(:page => params[:page])
+    else
+      @orphans = Orphan.paginate(:page => params[:page])
+    end
   end
```
razorcd: can't we extract this in a separate method?
maybe put it in the Model(helper) since the `.filter`(will be) and `.paginate` is also defined on the Model.

Then in the controller we can just have `Orphan.orphan_sort(...).paginate(...)`
! If there is no sort, `orphan_sort` should just return `self` otherwise it should just return ActiveRelation. (same as `.order(..)`

robbiejackson: yes I think that would be more elegant.
What file do you suggest? /models/orphan.rb ?

robbiejackson:
Column sort functionality moved from hq orphan controller into Orphan model - for comparison with similar functionality in the Sponsor area.

PR #382, opened 2015-05-26
OSRA415 - orphans#index Filters

gmunumel:
https://osraav.atlassian.net/browse/OSRA-415


gmunumel:
screenshot from 2015-05-26 19 39 32

razorcd:
please update the PR with the tests so I can take a look as discussed in last scrum. If still needed.

gmunumel:
Added the tests. They are failling. 

I've tested the functionality, the only field that is not working is `original_address_city` because I need to do a joins between `orphans` and `address` but I don't see how to do it. 

When I get the above done I'm going to make progress with the failling tests.

razorcd:
@gmunumel please review/merge:  https://github.com/gmunumel/osra/pull/3

PR #384, opened 2015-05-30
osra420: Sponsor table header sort

razorcd:
https://osraav.atlassian.net/browse/OSRA-420

Added class method `custom_sort` on Sponsor using a reusable concern
Added a clean chain of `Sponsor.filter.column_sort.paginate` to controller
Changed Sponsor index table headers to use links for sorting by column
Refactored `.where_with_conditions` to allow chaining even on empty arguments
Created partial `_sortable_table_header_link.html.erb` for adding sortable table header links
Added integration tests for sponsor index table

PR #385, opened 2015-05-31
OSRA419 sort table headers hq_partners

evo2013:
OSRA-419: 
1) Implemented sortable headers . Last column Province is sorting by province_id not province name. Province name does not exist in Partner's model & calling a separate tabel throws error. Your comments?
2) Implemented up/down arrows for sorting but partners view is not picking up assets/images .gif files. Your comments?
3) Working on adding sort tests.


razorcd:
First look at this PR: https://github.com/AgileVentures/osra/pull/384/files
It has the sorting done for the Sponsors table with helper methods that can be used for other Models also. Once this is merged feel free to apply it to Partners.

Sorting on a model relation is done using the `joins` method:
`Partner.joins(:province).order("provinces.name desc")`
But it would be better to use the `scope :column_sort` and refactor it to also accept relations.

Instead of the up/down arrows, better use the glyphicons from bootstrap. See the sponsors sorting PR, it has 2 css classes that be added to a `span` element.

evo2013:
Cristian, 
   Thanks for the feedback and I acknowledge receipt of this message. I just checked this email account and saw this message. I'll be working on the changes & tests this week and shall an update status by the weekend.

PR #386, opened 2015-06-06
OSRA425 Sponsor integration tests

achock:
Create sponsor integration test

### DIFF
--- a/spec/features/hq/create_sponsor_spec.rb
+++ b/spec/features/hq/create_sponsor_spec.rb
@@ -1,0 +1,37 @@
+require 'rails_helper'
+
+RSpec.feature 'User enters new sponsor data', :type => :feature do
+
+  background do
+    i_sign_in_as_admin
+  end
+
+  scenario 'User clicks the New Sponsor button' do
+    when_i_register_new_sponsor
+    and_i_click_button 'Create Sponsor'
+
+    then_the_sponsor_should_be_saved
+    and_i_should_be_on :new_sponsor_page
+  end
+
+  def when_i_register_new_sponsor
+    visit new_hq_sponsor_path
+    t = Time.now + 5.days
+    mDate = t.strftime("%Y-%m-%d")
+    fill_in 'Name', with: 'Jack Napier'
+    select 'Active', from: 'sponsor[status_id]'
+    select 'Male', from: 'sponsor[gender]'
+    fill_in 'Start date', with: '2015-06-01'
+    fill_in 'Requested orphan count', with: '1'
+    select 'Organization', from: 'sponsor[sponsor_type_id]'
+    select 'أهل الغربة وقت الكربة', from: 'sponsor[organization_id]'
+    select '(Afghanistan) أفغانستان', from: 'sponsor[country]'
+    select '**Add New**', from: 'sponsor[city]'
+    select 'Trisha Marquardt 1', from: 'sponsor[agent_id]'
+  end
+
+  def then_the_sponsor_should_be_saved
+    expect(page).to have_content 'Sponsor successfuly created'
+    expect(Sponsor.last.name).to eq 'Jack Napier'
+  end
+end
--- a/spec/features/hq/create_sponsor_spec.rb
+++ b/spec/features/hq/create_sponsor_spec.rb
@@ -4,14 +4,14 @@ RSpec.feature 'User enters new sponsor data', :type => :feature do
 
   background do
     i_sign_in_as_admin
+    an_agent_exists_with_username
   end
 
   scenario 'User clicks the New Sponsor button' do
     when_i_register_new_sponsor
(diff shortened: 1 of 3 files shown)


spec/features/hq/create_sponsor_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  scenario 'With invalid input' do
+    visit new_hq_sponsor_path
+    click_button 'Create and Add Another'
+    expect(page).to have_content 'Please fix the errors below and resubmit the form.'
+  end
+
+  def when_i_register_new_sponsor
+    FactoryGirl.create(:sponsor, city: 'Sigurdland')
+    visit new_hq_sponsor_path
```
razorcd: instead of going directly to new sponsor page, better go to sponsor index page and click New Sponsor button so we also that that button.
(no need to text the left navigation bar here because there is a different task for that)

spec/features/hq/create_sponsor_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  scenario 'With invalid input' do
+    visit new_hq_sponsor_path
+    click_button 'Create and Add Another'
+    expect(page).to have_content 'Please fix the errors below and resubmit the form.'
+  end
+
+  def when_i_register_new_sponsor
+    FactoryGirl.create(:sponsor, city: 'Sigurdland')
```
razorcd: I guess this record is created only to have a city in the DB?

I think we should skip `FactoryGirl.create` here and use `New city name` text field.(this will create a new city)
Fill in all fields.
Then click `Create and Add Another` and fill in the form with a second sponsor and use the same city by selecting the selectors option.
Then click "Create Sponsor" and check that it redirects and shows the correct sponsor and details (all details).
Check that you are on sponsor_show and see "Sponsor successfuly created"
Check all field are correct.
Then go to index and check that both sponsors are present.

1st sponsor should use Organization, 2nd sponsor should use Branch
Also we should fill in all fields, not only the required.

spec/support/feature_helpers/agent.rb
```diff
@@ -0,0 +1,12 @@
+module FeatureHelpers
+  module AgentHelper
+
+    def an_agent_exists
```
gmunumel: Why is there a difference between an agent and an agent with username?

achock: This is because when I tried to select certain agent from the dropdown I usually got a not found error as there was empty data pulled from the test_db. So Cristian had recommended to do a helper in which an agent was created with a username. This enabled me to select an agent an move along in the happy trail of testing.

spec/features/hq/create_sponsor_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    t = Time.now + 5.days
+    mDate = t.strftime("%Y-%m-%d")
+    fill_in 'Name', with: 'Jack Napier'
+    select 'Active', from: 'sponsor[status_id]'
+    select 'Female', from: 'sponsor[gender]'
+    fill_in 'Start date', with: mDate
+    fill_in 'Requested orphan count', with: '1'
+    select 'Organization', from: 'sponsor[sponsor_type_id]'
+    select 'أهل الغربة وقت الكربة', from: 'sponsor[organization_id]'
+    select '(Afghanistan) أفغانستان', from: 'sponsor[country]'
```
razorcd: maybe it's better to select an option that is not the first in the list?

spec/features/hq/create_sponsor_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  background do
+    i_sign_in_as_admin
+    an_agent_exists_with_username
+  end
+
+  scenario 'User clicks the New Sponsor button' do
+    when_i_register_new_sponsor
+
+    then_the_sponsor_should_be_saved
+    and_i_should_be_on :new_sponsor_page
```
razorcd: How about something like this:
```
#new
when_i_fill_in_new_sponsor_form_for "Paul"    #when_i_register_new_sponsor
click_button 'Create and Add Another' 
and_i_should_be_on :new_sponsor_page
then_the_sponsor_should_be_saved_for "Paul"   (not sure if this check should be part of integration tests)

when_i_fill_in_new_sponsor_form_for "Tom"
click_button 'Create'
then_the_sponsor_should_be_saved_for "Tom"   (not sure if this check should be part of integration tests)
and_i_should_be_on :sponsor_page, "Tom"
and_i_should_see_details_for "Tom"

#index
visit :sponsors_path
and_i_should_see "Paul"
and_i_should_see "Tom"
```

something like this will exercise `create and add another`, `create`, show view and index view

Also we have to add the `#edit` here too: Click Edit for Paul, change some fields (all would be best) and check show view for changed fields. It might be better to put `#edit` in a different scenario to keep them smaller.

PR #387, opened 2015-06-08
OSRA-432 Added integration tests for page navigation

ZmagoD:


razorcd:
I think this should only test the navigation and nothing else.

something like this:

First add `#main_nav_bar` id to `.navigation-left` nav element so it's unique.
Then:

```
scenario 'user clicks on Dashboard nav link' do
  whitin "#main_nav_bar" do
     click "Dashboard"
  end
    
   expect(page.current_path).to eq hq_dashboard_path
end
```
and same for each button.

The content of each page is tested separately. Including the logout.

ZmagoD:
Done, but for the "OSRA" link I left a test commented for now beacuse now the root render acive admin.

spec/features/hq/navigation_menu.spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    
+  background do
+    i_sign_in_as_admin
+  end
+  scenario 'user cilco on OSRA link' do
+    within "#main_nav_bar" do
+      click_link "OSRA"
+    end
+   # expect(page).to eq hq_index_path
+  end
```
razorcd: better add a `pending` here instead of only a comment so we know to come back when we have the correct path.
when you run `rspec` the pending tests are displayed at the end in orange/yellow

spec/features/hq/navigation_menu.spec.rb
```diff
@@ -0,0 +1,57 @@
+require 'rails_helper'
+require 'spec_helper'
+
+RSpec.feature 'Navigation menu', :type => :feature do
+    
+  background do
+    i_sign_in_as_admin
+  end
+  scenario 'user cilco on OSRA link' do
```
razorcd: typo: clicks

razorcd:
:+1:

gmunumel:
:shipit:

PR #399, opened 2015-08-27
Fix randomly failing test in orphans form spec

PurityControl:
- cast the 5 digit faker guardian_id_num to an integer to prevent
  leading zeros cause intermittent failures for values set on input fields.
- this fixes a previous fail when running :-
    `rspec --order random:28201`

### DIFF
--- a/spec/views/hq/orphans/_form_spec.rb
+++ b/spec/views/hq/orphans/_form_spec.rb
@@ -44,12 +44,7 @@ RSpec.describe "hq/orphans/_form.html.erb", type: :view do
      "guardian_name", "guardian_relationship",
      "guardian_id_num", "contact_number", "alt_contact_number"].each do |field|
       if orphan_full[field]
-        if !expect(rendered).to have_selector("input[id='orphan_#{field}'][value=\"#{orphan_full[field].to_s}\"]")
-          show_me_redered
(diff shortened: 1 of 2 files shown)


spec/views/hq/orphans/_form_spec.rb
```diff
@@ -44,12 +44,7 @@ def render_orphan_form current_orphan
      "guardian_name", "guardian_relationship",
      "guardian_id_num", "contact_number", "alt_contact_number"].each do |field|
       if orphan_full[field]
-        if !expect(rendered).to have_selector("input[id='orphan_#{field}'][value=\"#{orphan_full[field].to_s}\"]")
-          show_me_redered
-          binding.pry
```
PurityControl: The `binding.pry` makes this look like it may have been a problem in the past.
However the `if` never gets evaluated because if the spec fails it bails out before the `if` is even called.

razorcd:
thanks @PurityControl 
This id_num error was bugging me for a while
:+1:

spec/views/hq/orphans/_form_spec.rb
razorcd: yes, you are right. That `if` will never catch it. :+1:

PR #401, opened 2015-08-28
OSRA-425, initial implementation of sponsor crud feature.

PurityControl:
- change sponsor saved message from "created" to "saved"
- implement sponsor crud feature

### DIFF
--- a/spec/features/hq/crud_sponsor_spec.rb
+++ b/spec/features/hq/crud_sponsor_spec.rb
@@ -1,0 +1,190 @@
+require 'rails_helper'
+require 'spec_helper'
+
+=begin
+Feature:
+    As a site admin for osra
+    So that I may maintain the information about osra sponsors
+    I can create, read and update sponsors in the admin section
+=end
+RSpec.feature 'CRUD Sponsor', :type => :feature do
+
+  background do
+    @sponsors = create_list :sponsor_full, 3
+    i_sign_in_as_admin
+  end
+
+  scenario  'There should be a list of partners on the admin index page' do
+    visit hq_sponsors_path
+    @sponsors.each do |sponsor|
+      and_i_should_see sponsor.name
+    end
+  end
+
+  scenario 'I should see the generated osra numbers on the admin index page' do
+    visit hq_sponsors_path
+    @sponsors.each do |sponsor|
+      expect(page).to have_content(sponsor.osra_num)
+    end
+  end
+
+  scenario 'I should see the required fields on the index page' do
+    visit hq_sponsors_path
+    @sponsors.each do |sponsor|
+      i_should_see_the_following_fields_on_the_page [
+        { field: 'osra_num',          value: sponsor.osra_num },
+        { field: 'name',              value: sponsor.name },
+        { field: 'status',            value: sponsor.status.name },
+        { field: 'start_date',        value: format_date(sponsor.start_date) },
+        { field: 'request_fulfilled', value: request_fulfilled_text(sponsor) },
+        { field: 'sponsor_type',      value: sponsor.sponsor_type.name },
+        { field: 'country',           value: en_ar_country(sponsor.country) }
+      ]
+    end
+  end
+
+  scenario 'I should see the required fields on the partner show page' do
+    sponsor = @sponsors.first
+    given_i_am_on_page_for_sponsor 'Show Sponsor', sponsor.name
+    i_should_see_the_following_fields_on_the_page [
+      { field: 'name',                   value: sponsor.name },
+      { field: 'osra_num',               value: sponsor.osra_num },
+      { field: 'status',                 value: sponsor.status.name },
+      { field: 'gender',                 value: sponsor.gender },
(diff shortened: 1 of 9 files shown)


spec/features/hq/crud_sponsor_spec.rb
```diff
@@ -0,0 +1,190 @@
...
+      and_i_should_see sponsor.name
+    end
+  end
+
+  scenario 'I should see the generated osra numbers on the admin index page' do
+    visit hq_sponsors_path
+    @sponsors.each do |sponsor|
+      expect(page).to have_content(sponsor.osra_num)
+    end
+  end
```
razorcd: I think `scenario` from L17 and L24 are covered in `scenario` L31

spec/features/hq/crud_sponsor_spec.rb
```diff
@@ -0,0 +1,190 @@
...
+        { field: 'status',            value: sponsor.status.name },
+        { field: 'start_date',        value: format_date(sponsor.start_date) },
+        { field: 'request_fulfilled', value: request_fulfilled_text(sponsor) },
+        { field: 'sponsor_type',      value: sponsor.sponsor_type.name },
+        { field: 'country',           value: en_ar_country(sponsor.country) }
+      ]
+    end
+  end
+
+  scenario 'I should see the required fields on the partner show page' do
```
razorcd: `sponsor show page` instead of `partner show page`

spec/features/hq/crud_sponsor_spec.rb
```diff
@@ -0,0 +1,190 @@
...
+    i_should_not_be_able_to_change 'Organization', 'sponsor'
+    i_should_not_be_able_to_change 'Branch', 'sponsor'
+  end
+
+  scenario "can't delete a sponsor from the sponsor show page" do
+    given_i_am_on_page_for_sponsor 'Show Sponsor', @sponsors.first.name
+    and_i_should_not_see 'Delete Sponsor'
+  end
+
+  scenario "returns to partner show page on cancelling edit partner" do
```
razorcd: `returns to sponsor show page on cancelling edit sponsor`

spec/features/hq/crud_sponsor_spec.rb
```diff
@@ -0,0 +1,190 @@
...
+    name = page_name.downcase
+    case name
+      when 'edit sponsor' then
+        edit_hq_sponsor_path id
+      when 'show sponsor' then
+        hq_sponsor_path id
+      when 'new sponsor' then
+        new_hq_sponsor_path
+    end
+  end
```
razorcd: maybe methods from L140 and L149 need to go to `osra/spec/support/feature_helpers/navigation.rb`?
Also `get_path` and `and_i_should_be_on` might need refactoring so they are helper methods and don't have expectations.

razorcd:
thanks @PurityControl for your help!
If you don't have time to do these small changes then I will do them next week.
I need to review all integration tests anyway, too make sure helper methods are used correctly everywhere.

spec/features/hq/crud_sponsor_spec.rb
PurityControl: I have pushed all the other changes suggested. However, this change probably needs further thought ...

I agree in principle but how do you plan on approaching this?
- `path_to_sponsor_role` could be incorporated into `get_path` if `get_path` had no expectations
- unless you rename `and_i_should_be_on` the method name gives the impression the method has an expectation
- If you remove the expectations and make them helpers, given that these expectations are needed, otherwise they wouldn't be there, where do you put them? Are you wanting to push them back to the scenario in question. If so how much does this break dry?
- `path_to` has a mixture of paths ending in path or page, perhaps we should use just one for consistency.
- `and_i_should_be_on` seems to require a single key hash to work, I can't put my finger on it but it doesn't sit easy with me. It seems a little brittle.

Regardless, it is probably worth refactoring before things get even more complicated.

Thoughts?

razorcd: If we move the expectation from `get_path` to the end of `and_i_should_be_on` it will do what the method says.
And you can use `get_path` instead of the new `path_to_sponsor_role` method, like you specified. 

This also implies checking all scenarios that use `get_path` and make sure an expectation is not necessary there (if expectation is necessary then it probably needs `and_i_should_be_on`).

Yes we should use only `page` in `get_path`.

Yes, that `field_hash.keys.first.to_sym` is brutal.
Maybe something like:
```
case field_hash.keys.map(&:to_sym)
  when [:sponsor_name] ...
  when [:sponsor_name, :user_name] ... #in future
```

PurityControl: Sounds like a plan.
`get_path` is only called in `navigation.rb` so fortune is smiling on us :smile:

PurityControl: changes pushed up for review.

razorcd:
:+1: looks much better

PR #402, opened 2015-09-12
Osra433 Integrations for uploading orphan lists

razorcd:
https://osraav.atlassian.net/browse/OSRA-433


PurityControl:
looks good to me

PR #404, opened 2015-09-13
End sponsorships feature #OSRA-438

PurityControl:
OSRA-438
- make inactive sponsorships have a css class of .inactive_sponsors_index
- add end sponsorship feature
- add delete sponsorship feature


razorcd:
:+1:

PR #412, opened 2016-07-13
OSRA-441 Display sponsor on orphan show page

Maroo-b:
@motasem-salem it's WIP pull request I just added a feature test but I want to confirm that an orphan could have only one sponsor ?


motasem-salem:
An orphan can have at most one active sponsor. He / she can have multiple sponsors who are not active, but we only display the active one (if available)

Maroo-b:
https://osraav.atlassian.net/browse/OSRA-441

- Add link to sponsor from orphan show page
sponsor_link

PR #415, opened 2016-07-27
Osra-444 add export functionality to orphan model

Maroo-b:
https://osraav.atlassian.net/browse/OSRA-444

- add export to csv functionality to orphans and sponsors

### DIFF
--- a/app/models/orphan.rb
+++ b/app/models/orphan.rb
@@ -1,3 +1,4 @@
+require 'csv'
 class Orphan < ActiveRecord::Base
 
   enum status: [
@@ -88,6 +89,8 @@ class Orphan < ActiveRecord::Base
   has_one :partner, through: :orphan_list, autosave: false
 
   delegate :province_code, to: :partner, prefix: true
+  delegate :province_name, to: :original_address
+  delegate :name, to: :partner, prefix: true
 
   accepts_nested_attributes_for :current_address, allow_destroy: true
   accepts_nested_attributes_for :original_address, allow_destroy: true
@@ -121,6 +124,17 @@ class Orphan < ActiveRecord::Base
 
   def current_sponsor
     current_sponsorship.sponsor if sponsored?
+  end
+
+  def self.to_csv(records = [])
+    attributes = %w(osra_num full_name father_name date_of_birth gender province_name partner_name 
+      father_is_martyr father_deceased mother_alive priority status sponsorship_status)
+    CSV.generate(headers: true) do |csv|
+      csv << attributes.map(&:titleize)
+      records.each do |orphan|
+        csv << attributes.map { |attr| orphan.public_send(attr) }
+      end
+    end
   end
 
 private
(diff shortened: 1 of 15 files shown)


app/models/orphan.rb
```diff
@@ -123,6 +126,17 @@ def current_sponsor
     current_sponsorship.sponsor if sponsored?
   end
 
+  def self.to_csv(records = [])
+    attributes = %w(osra_num full_name father_name date_of_birth gender province_name partner_name 
+      father_is_martyr father_deceased mother_alive priority status sponsorship_status)
+    CSV.generate(headers: true) do |csv|
+      csv << attributes.map(&:titleize)
+      records.each do |orphan|
+        csv << attributes.map { |attr| orphan.send(attr) }
```
b264: `public_send` might be better here than `send` which also allows for calls to private/protected methods

app/models/orphan.rb
```diff
@@ -132,7 +132,7 @@ def self.to_csv(records = [])
     CSV.generate(headers: true) do |csv|
       csv << attributes.map(&:titleize)
       records.each do |orphan|
-        csv << attributes.map { |attr| orphan.send(attr) }
+        csv << attributes.map { |attr| orphan.public_send(attr) }
```
Maroo-b: @b264 I updated the PR to use public_send instead, thx for the review :)

b264:
I'm not sure about that test failure... I don't know if hardcoding the country is the best fix.  But I don't know what the best fix would be.  This works, but I'm not sure if it's ideal to use in the test or not.  This would seem to be an implementation detail of the CSV output, I think.
``` ruby
2.3.0 :001 > YAML.parse('"foo"').first.value== YAML.parse('foo').first.value
 => true```
@razorcd @motasem-salem any ideas?

File: app/models/orphan.rb (hunk 2 above)
Comment (b264) [inline app/models/orphan.rb]: :-)

File: spec/models/sponsor_spec.rb [hunk 3]
```diff
@@ -368,6 +368,15 @@
       it 'returns only orphans that are currently sponsored' do
         expect(new_sponsor.currently_sponsored_orphans).to match_array [current_orphan]
       end
+
+      describe '#to_csv' do
+        let!(:sponsor) {build(:sponsor, name: 'John Doe', gender: 'Male', requested_orphan_count: 4, active_sponsorship_count: 1, request_fulfilled: false, country: 'FR')}
+        it 'generates csv content for given sponsors' do
+          output = "Osra Num,Name,Status,Start Date,Request Fulfilled,Sponsor Type,Country\n,John Doe,#{sponsor.status.name},#{sponsor.start_date},No (1/4),#{sponsor.sponsor_type.name},#{en_ar_country(sponsor.country)}\n"
```
b264:
``` ruby
output= /Osra Num,Name,Status,Start Date,Request Fulfilled,Sponsor Type,Country\n,John Doe,#{Regexp.escape sponsor.status.name},#{Regexp.escape sponsor.start_date},No \(1\/4\),#{Regexp.escape sponsor.sponsor_type.name},[\"]?#{Regexp.escape en_ar_country(sponsor.country)}[\"]?\n/m
expect(output.match Sponsor.to_csv([sponsor])).to be
``` might quash that test failure without hardcoding the country, to allow that part to be tested too

b264:
:+1:

Maroo-b:
@b264 thanks for the review that was really helpful.

b264:
:shipit:

PR #417, opened 2016-08-16
Osra 447 export csv with filters

develaper:
Related to https://osraav.atlassian.net/browse/OSRA-447

- Add filters, sort direction and sort column if exists to the 'export to csv' link in the view and in the controller for orphans and sponsors.

### DIFF
--- a/app/controllers/hq/orphans_controller.rb
+++ b/app/controllers/hq/orphans_controller.rb
@@ -9,17 +9,17 @@ class Hq::OrphansController < HqController
     @current_sort_direction = valid_sort_direction
 
     @filters = filters_params
-    @orphans = Orphan
+    @orphans_before_paginate = Orphan
       .filter(@filters)
       .includes(valid_sort_columns_included_resource)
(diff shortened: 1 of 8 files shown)


b264:
The code looks good but I don't see any test that covers the essence of OSRA-447 which is to verify the export takes current filters into account.  Maybe we can add a test or two to cover when filters are active?

develaper:
To be completely honest, I wasn't very sure about how to assign values and test it. Any advice or improvement is welcome.

b264:
I would maybe add a test inside `RSpec.describe Hq::OrphansController` `describe '#index'` like `specify "filter-CSV"` and verifies one orphan's name exists in the output, and another's does not, based on whichever filter you put in the spec and two named orphans you create for the test, and also the same sort of thing for `RSpec.describe Hq::SponsorsController`.  Currently I'm not seeing any tests on the controller CSV formats.  (Did I miss them?)  The view specifications look great and there are CSV specs on the models already.

b264:
:+1:

Maroo-b:
@motasem-salem please review the updates ?

PR #418, opened 2016-08-24
Osra 448 display total number of records in list pages

develaper:
Related to : https://osraav.atlassian.net/browse/OSRA-448

- Add total number of sponsors and orphans in each index page based in the message displayed in the admin version.

osra-448

### DIFF
--- a/spec/views/hq/sponsors/index_spec.rb
+++ b/spec/views/hq/sponsors/index_spec.rb
@@ -1,4 +1,5 @@
 require 'rails_helper'
+require 'will_paginate/array'
 
 RSpec.describe 'hq/sponsors/index.html.haml', type: :view do
   before :each do
@@ -7,6 +8,13 @@ RSpec.describe 'hq/sponsors/index.html.haml', type: :view do
     assign(:sort_by, {})
     assign(:sortable_by_column, true)
   end
+
+  let(:sponsors_count) { build_stubbed_list(:sponsor, 5).count }
+  let(:sponsors) do
+    items = build_stubbed_list :sponsor, 5
+    items.paginate(:page => 2, :per_page => 2)
+  end
+
   it 'should delegate to partial' do
     render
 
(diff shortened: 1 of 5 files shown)


spec/views/hq/sponsors/index_spec.rb
```diff
@@ -1,4 +1,5 @@
 require 'rails_helper'
+require 'will_paginate/array'
```
b264: I think the alternative to this is to replace `assign(:sponsors, [])` with `assign(:sponsors, Sponsor.none)` which is actually closer to the conditions the actual code will encounter, as it is an empty ActiveRecord::Relation enumerable instead of an Array

b264: Or maybe `Class.new(ActiveRecord::Base).none`

b264: This actually appears in the code in many places, but I think phasing it out each time it's encountered is a good idea...

motasem-salem:
Can we move the count to the top of the page, maybe just below "Currently sponsoring..."

develaper:
Should we do the same with the rest of the lists? I mean "Users", "Partners" and  "Admin users" or there is any reason to avoid it?

spec/views/hq/sponsors/index_spec.rb
develaper: Removing this line for me results in this error message:
`Failure/Error: items.paginate(:page => 2, :per_page => 2)
     NoMethodError:
       undefined method `paginate' for #<Array:0x0000000baa0838>`
Any idea?

b264: You have to find the Array and replace it with a relation.  So somewhere, instead of `[]` or something like that, it needs to be `Sponsor.none` or similar

b264: Sorry, that must be `build_stubbed_list` that's doing that.  I'd leave this here then.

b264: :+1:

b264:
@develaper If so, I think those would be different PRs.

b264:
:+1:  :shipit:

PR #419, opened 2016-10-26
Include active sponsor

NikitaAvvakumov:
- add display of current sponsor, if any, to orphans index view & exported CSV
- fix N+1 queries
- refactor query in OrphansController#index

This PR breaks a number of unit tests, mostly controller & view. All are low-level tests that are very fragile as they are tied too closely to the implementation detail. Rather than fixing them, I am inclined to either improve them or to drop them entirely. Would prefer to handle that separately in order to deliver the feature to the client ASAP


NikitaAvvakumov:
@b264 Fixed the tests in place.

Going forward, refactoring and/or removing many of these should be a priority to avoid ~doubling the workload for each change or new feature. Right now, I'm tempted to just `rm -rf spec/controllers`. Back in the day, we drank way too much textbook Kool-Aid and chased after complete coverage instead of thinking about the purpose of these tests

b264:
:-D

b264:
:shipit:

Maroo-b:
+1

PR #422, opened 2016-11-03
Update CONTRIBUTING

b264:
Add server dns


CONTRIBUTING.md
```diff
@@ -155,3 +155,12 @@ into any problems during this process, please reach out to the OSRA team on
...
 We look forward to having you on our team!
+
+
+#### staging
+
+The test server is at http://osra-new-staging.herokuapp.com/
+
+#### production
+
+The production server is at http://osra.herokuapp.com/
```
NikitaAvvakumov: For the sake of completeness, there is also the dev server at osra-testing.herokuapp.com. That said, I'm not sure how useful this ends up being as all apps require login, and we are probably not going to post credentials here.

PR #423, opened 2016-11-03
Force SSL in production

NikitaAvvakumov:
Nothing more should be needed - all *.herokuapp.com domains are covered by Heroku's SSL cert


b264:
:-D

PR #425, opened 2016-11-05
Remove aa

NikitaAvvakumov:
☢️ 💣 💥  => 🎉


razorcd:
:+1:

razorcd:
should we keep a redirect route from `/hq` to `/` ? Just so some old links keep working.

NikitaAvvakumov:
@razorcd Good idea - added a globbed redirect for any `GET` requests to the 'hq' namespace ([commit](https://github.com/AgileVentures/osra/pull/425/commits/aaae0e598dc72f6b05720f3ef0d562fe54798548)). Should work well enough until muscle memory and bookmarks get updated.

Another thought I had was to create a Rack middleware to strip `/hq` from all request paths - should be possible, but I don't think it's really necessary.

b264:
:astonished:

NikitaAvvakumov:
Discovered a problem after deploying this to Heroku: the `filter` scope defined on `Orphan` in the `OrphanAttrFilter` concern now raises `NoMethodError (undefined method `filter' for #<Orphan::ActiveRecord_Relation:0x007f5a70800698>)` in `OrphansController#index`. Same situation with `Sponsor.filter`. Locally, I didn't see this in the browser, but `spec/models/concerns/orphan_attr_filter_spec.rb` was failing with the same exception until I converted it to an `RSpec.describe Orphan, type: :model`. Strangely, there were no such problems with `Sponsor.filter`.

In the `OrphanAttrFilter` concern I could define an instance method that would be available on `Orphan` objects, and it was the `filter` scope specifically that wasn't getting picked up.

Will keep investigating, but any pointers appreciated

NikitaAvvakumov:
[This commit](https://github.com/AgileVentures/osra/pull/425/commits/d0445006c067cfddb8cef15404f68ce4c2bfbccd) fixes the problem mentioned above, though I'm not entirely certain as to why. The issue reproducibly showed up with `RAILS_ENV=production` and did not with `RAILS_ENV=development`.

This branch is deployed to https://osra-nga.herokuapp.com, you can use the same credentials as for osra-new-staging. Ping me in Slack if needed

b264:
`Missing functionality:`
Partners index is missing the Download CSV link (is this used/needed?)
Users index is missing the Download CSV link (is this used/needed?)
Admin Users index is missing the Download CSV link (is this used/needed?)

`Reduced functionality:`
`No route matches [GET] "/admin/sponsors"` for example on previously bookmarked pages, I think we should also redirect these
Orphans index missing colour highlighting on boolean values
Sponsors index missing colour highlighting on boolean values

`Nice touches:`
The `OSRA` button in the nav bar is no longer needed

b264:
The `NoMethodError` on production only was probably caused by preloading, I have seen similar things before.  Sometimes you can add `require` statements in the non-Rails Ruby style to fix that

b264:
`pending_orphan_lists/upload.html.erb ` seems to not be reachable

app/controllers/application_controller.rb
```diff
@@ -2,4 +2,16 @@ class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
+
+  NAVIGATION_BUTTONS= [
+    { text: 'OSRA', href: 'root_path', path_regex: /^\/$/, glyph: 'glyphicon-home' },
```
b264: I don't think this button adds any functionality beyond the `Dashboard` button

app/admin/partner.rb
```diff
@@ -1,74 +0,0 @@
...
-      f.input :contact_details
-    end
-    f.actions do
-      f.action :submit
-      f.action :cancel, :label => "Cancel", :wrapper_html => { :class => "cancel" }
-    end
-  end
-
-  action_item :upload_orphan_list, only: :show do
-    link_to('Upload Orphan List', upload_admin_partner_pending_orphan_lists_path(partner)) if partner.active?
```
b264: I don't see this reachable any more, does the client still need this or was it only for migration to the new management system?

app/controllers/application_controller.rb
NikitaAvvakumov: Removed. Neither it nor the "Dashboard" button have any value at the moment, but I'd like to still hold on to the idea of a functional dashboard with notifications etc., so it gets to stay :)

app/admin/partner.rb
NikitaAvvakumov: `partners/<partner_id>/pending_orphan_lists/upload` is reachable via the partner show view

app/controllers/application_controller.rb
b264: :laughing:

b264: I'd argue the Dashboard button does hold value, if you are at a business with confidential information on your screen and someone comes to talk to you, you can click it to show an OSRA logo instead of confidential information

NikitaAvvakumov:
- re: missing CSV export options for partners, users & admin users: Not needed as far as I understand. Overall, I'm acting under the assumption that the client has been using the `hq` namespace exclusively for a while now and it is feature-equivalent to the `admin` namespace. I've added redirection for `/admin` routes, but the hope is that they haven't been used for some time anyway. Perhaps all this is worth confirming

- re: missing colour highlighting on booleans: basically, as above - if the client hasn't asked for it, we let it be. Particularly in the `hq` UI, highlighting would really stand out - unless there is a real benefit to quickly distinguishing between the 'yes' and 'no' in those categories, colour might actually be disruptive to the UX

b264:
I think we should get confirmation from the client then that they aren't using the export functions on those pages?  Or alternatively we could dig into the server log and see if they've been accessed...

app/admin/partner.rb
b264: :+1:

b264:
:shipit:

PR #426, opened 2016-11-07
Update projects gems

Maroo-b:
Addresses: https://osraav.atlassian.net/browse/OSRA-445
update gems:
- Devise
- Uglifier
- jqery-rails


b264:
:shipit:

PR #427, opened 2016-11-09
Allow overriding db settings

b264:
This allows a developer to use their own database password instead of the ones hardcoded into the project


b264:
For some reason this built on Travis instead of semaphore https://travis-ci.org/b264/osra/jobs/174374245

PR #428, opened 2016-11-09
Allow overriding db settings

b264:
This allows a developer to use their own database password instead of the ones hardcoded into the project


b264:
example
```shell
devbox@ubuntu-server-xenial:~/code/osra$ DB_PASSWORD=db_password DB_USER=db_user rails server
```

PR #429, opened 2016-11-16
Include current sponsor's OSRA num in orphan index

NikitaAvvakumov:
Client's feedback on including current sponsor info in orphan index view & CSV export:
> The functionality  is working fine 
Is it possible to add the sponsor's OSRA number to the report ?


spec/models/orphan_spec.rb
```diff
@@ -437,9 +437,10 @@
...
-            orphan_attrs.merge!({ partner_name: "partner name" })
+            orphan_attrs[:partner_name] = "partner name"
+            orphan_attrs[:current_sponsor_for_csv] = []
             orphan = OpenStruct.new(orphan_attrs)
-            output = "Osra Num,Full Name,Father Name,Date Of Birth,Gender,Province Name,Partner Name,Father Is Martyr,Father Deceased,Mother Alive,Priority,Status,Sponsorship Status,Current Sponsor\n,John Mark Doe,Mark Doe,#{orphan.date_of_birth},Male,#{orphan.province_name},partner name,#{orphan.father_is_martyr},#{orphan.father_deceased},#{orphan.mother_alive},#{orphan.priority},#{orphan.status},#{orphan.sponsorship_status},--\n"
+            output = "Osra Num,Full Name,Father Name,Date Of Birth,Gender,Province Name,Partner Name,Father Is Martyr,Father Deceased,Mother Alive,Priority,Status,Sponsorship Status,Current Sponsor,Sponsor OSRA Num\n,John Mark Doe,Mark Doe,#{orphan.date_of_birth},Male,#{orphan.province_name},partner name,#{orphan.father_is_martyr},#{orphan.father_deceased},#{orphan.mother_alive},#{orphan.priority},#{orphan.status},#{orphan.sponsorship_status}\n"
```
b264: should this end with `#{orphan.sponsorship_status},,\n"` or should we add a test where there is a sponsor present.  I'd have to check the csv specs

NikitaAvvakumov: `current_sposor_for_csv` returns an empty array if there is no sponsor, so no commas. I wouldn't test this additionally with a sponsor as it really would just be testing Ruby's array concatenation.

b264: That's the things, each row of the csv should have the same number of commas, right?  So we should be testing that they indeed have that.  The header row has the same number as the data row

b264: I think the only reason it might work is because they are at the end, but what if they were eliminated in the middle, like
```
fooA,fooB,fooC\n
1,2,3\n
4,6\n
7,8,9\n
```

b264: Which column does 6 go in?  `fooB` or `fooC`

b264: If it's to spec, I'm okay with omitting them as long as we know we are doing that and it's not an accident.

NikitaAvvakumov: @b264 @Maroo-b You make a good point - the current implementation takes too much advantage of the fact that the sponsor is in the last 2 columns. Should we decide to add another, the sponsor CSV logic will have to be changed as well.

I'll fix this

Maroo-b:
@NikitaAvvakumov that was a simple and clever fix :+1: 

@b264 should we merge this PR ?

b264:
:+1:

PR #430, opened 2016-11-18
Adult orphan display

NikitaAvvakumov:
> As an OSRA orphan manager
So that I can quickly identify orphans who are above 18 years old
I need the rows of such orphans to be highlighted in red on the index view

~Waiting for client to confirm whether orphans are considered adults at 18 or 19 years of age.~

Client responds:
> orphan becomes adult on his/her 18th birthday

orphans


b264:
:+1: :shipit:

PR #433, opened 2016-11-29
No age highlight for inactive orphans

NikitaAvvakumov:
Client request:
> The orphan which is inactive doesn't have to be highlighted

### DIFF
--- a/spec/helpers/orphan_helper_spec.rb
+++ b/spec/helpers/orphan_helper_spec.rb
@@ -1,7 +1,7 @@
 require "rails_helper"
 
 RSpec.describe OrphanHelper, type: :helper do
-  describe "#active_adult?" do
+  describe "#orphan_highlight_class" do
     context "when active child" do
       example do
         orphan = double "orphan", adult?: false, active?: true
--- a/spec/helpers/orphan_helper_spec.rb
+++ b/spec/helpers/orphan_helper_spec.rb
@@ -6,7 +6,7 @@ RSpec.describe OrphanHelper, type: :helper do
       example do
         orphan = double "orphan", adult?: false, active?: true
 
-        expect(active_adult?(orphan)).to be_nil
+        expect(orphan_highlight_class(orphan)).to be_nil
       end
     end
 
@@ -14,7 +14,7 @@ RSpec.describe OrphanHelper, type: :helper do
       example do
         orphan = double "orphan", adult?: false, active?: false
 
-        expect(active_adult?(orphan)).to be_nil
+        expect(orphan_highlight_class(orphan)).to be_nil
       end
     end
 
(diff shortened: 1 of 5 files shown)


spec/helpers/orphan_helper_spec.rb
```diff
@@ -1,20 +1,36 @@
...
-    context "when orphan is a child" do
-      it "returns 'child'" do
-        orphan = double "orphan", age_in_years: 17
+  describe "#active_adult?" do
+    context "when active child" do
+      example do
+        orphan = double "orphan", adult?: false, active?: true
 
-        expect(child_or_adult? orphan).to eq "child"
+        expect(active_adult?(orphan)).to be_nil
```
b264: I think asserting `nil` might be overkill.  Should we just check for falsey value?

spec/helpers/orphan_helper_spec.rb
```diff
@@ -1,20 +1,36 @@
...
-        expect(child_or_adult? orphan).to eq "adult"
+        expect(active_adult?(orphan)).to be_nil
+      end
+    end
+
+    context "when inactive adult" do
+      example do
+        orphan = double "orphan", adult?: true, active?: false
+
+        expect(active_adult?(orphan)).to be_nil
```
b264: Same here about `nil` versus any falsey object

Maroo-b: @b264 good point, by convention a method that ends with a question mark is a predicate method (which returns true or false) but in our case `active_adult?` returns either 'active_adult' (the css class name) or nil.
I guess that renaming the method `active_adult?` to `highlight_class` will convey the meaning that we are using a dynamic css class, and also we can check nil value explicitly.

b264: Even returning `'active_adult'` or `nil` would technically still test fine as truthy and falsey values, I'm merely suggesting we don't specifically test for `nil` or `false` or `true`

NikitaAvvakumov: Agreed that the method name was poorly chosen. Renaming it to `orphan_highlight_class` with no other changes works the best for me. In this case, I think testing for a `nil` return is the right thing to do as we want to make sure that the method does not return any CSS class unless the conditions are met.

spec/helpers/orphan_helper_spec.rb
Maroo-b: @NikitaAvvakumov it seems that we should update the method naming in helper spec to make the build pass

NikitaAvvakumov: @Maroo-b 🤕 This is what happens when you code in 30-second increments...

PR #439, opened 2016-12-12
Payments model

b264:
https://osraav.atlassian.net/browse/OSRA-452


Maroo-b:
@b264 great start, I suggest to change '﷼‎' to 'SAR' because in Arabic we write from right to left (it looks awkward to me)

b264:
@Maroo-b So how do you think is best?  I'm unfamiliar with their currency.  This is also why I made a class to do this and only this :stuck_out_tongue:
`expect(Halalas.new(-1234).to_s).to eq 'SAR (12.34)‎'`

Maroo-b:
+1

db/migrate/20161212003047_add_payments.rb
```diff
@@ -0,0 +1,10 @@
+# coding: utf-8
+# frozen_string_literal: true
+class AddPayments < ActiveRecord::Migration
+  def change
+    create_table :payments do |t|
+      t.integer :amount, null: false
```
b264: using integers here (halala) specifically avoids things like https://bugs.ruby-lang.org/issues/12958 and all machine epsilon errors in general

Maroo-b: that was very interesting to know, I agree that using integers for calculation than  dividing by 100 for display is the best choice. I had done the mistake of using decimal type for currency in an old project :cry:

app/helpers/halalas.rb
```diff
@@ -0,0 +1,24 @@
...
+    @riyals= Rational(@halalas, 100)
+  end
+
+  def to_s
+    riyals_string + CURRENCY
+  end
+
+private
+
+  def riyals_string
```
PurityControl: Consider blurring the line between data and behaviour. When data is accessed by message sending it insulates you from future changes.

```
attr_reader :riyals

def riyals_string
  if riyals < 0
    '(%.2f)' % riyals.abs
  else
     '%.2f' % riyals
   end
end
```

b264: Good point. Thanks!

app/helpers/halala.rb
```diff
@@ -0,0 +1,25 @@
+# coding: utf-8
+# frozen_string_literal: true
+class Halala
+  CURRENCY= ' SAR'
+  attr_reader :riyal
```
PurityControl: This is subtly different from your previous behaviour. If you are happy for riyal to be accessible outside the class then this is fine. If you would rather it was kept confined to the class, as before, then it needs to be private.

b264: I'll make it private for now, if we need it later, we can wrap its functionality with tests

app/models/payment.rb
```diff
@@ -0,0 +1,5 @@
+# coding: utf-8
+# frozen_string_literal: true
+class Payment < ActiveRecord::Base
```
NikitaAvvakumov: Could be useful to document the requirements & usage of `amount` here. I think it's necessary to make it very discoverable that the value must be a whole number of halalas ("cents") rather than a fractional representation of riyals ("dollars")

PR #441, opened 2017-01-03
Add cashbox model [OSRA-451]

Maroo-b:
- Addresses: https://osraav.atlassian.net/browse/OSRA-451

### DIFF
--- a/app/models/cashbox.rb
+++ b/app/models/cashbox.rb
@@ -2,4 +2,6 @@ class Cashbox < ActiveRecord::Base
   has_many :deposits, class_name: 'Payment', foreign_key: 'destination_id'
   has_many :withdrawls, class_name: 'Payment', foreign_key: 'source_id'
   belongs_to :cashboxable, polymorphic: true
+
+  validates :balance, numericality: { only_integer: true }, allow_nil: false
 end
--- a/app/models/cashbox.rb
+++ b/app/models/cashbox.rb
@@ -1,0 +1,5 @@
+class Cashbox < ActiveRecord::Base
+  has_many :deposits, class_name: 'Payment', foreign_key: 'destination_id'
+  has_many :withdrawls, class_name: 'Payment', foreign_key: 'source_id'
+  belongs_to :cashboxable, polymorphic: true
+end
--- a/app/models/cashbox.rb
+++ b/app/models/cashbox.rb
@@ -1,7 +1,10 @@
 class Cashbox < ActiveRecord::Base
   has_many :deposits, class_name: 'Payment', foreign_key: 'destination_id'
-  has_many :withdrawls, class_name: 'Payment', foreign_key: 'source_id'
+  has_many :withdrawals, class_name: 'Payment', foreign_key: 'source_id'
   belongs_to :cashboxable, polymorphic: true
 
-  validates :balance, numericality: { only_integer: true }, allow_nil: false
+  def total
+    deposits.sum(:amount) - withdrawals.sum(:amount)
+  end
+
 end
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -5,5 +5,8 @@
 # "amount" is in integer quantity in halalas, or hundredths of a riyal.
 
 class Payment < ActiveRecord::Base
+  belongs_to :source, class_name: 'Cashbox'
+  belongs_to :destination, class_name: 'Cashbox'
+
   validates :amount, numericality: { only_integer: true }, allow_nil: false
 end
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -8,5 +8,6 @@ class Payment < ActiveRecord::Base
   belongs_to :source, class_name: 'Cashbox'
   belongs_to :destination, class_name: 'Cashbox'
 
-  validates :amount, numericality: { only_integer: true }, allow_nil: false
+  validates :amount, numericality: { only_integer: true, greater_than: 0 },
+    allow_nil: false
 end
--- a/db/migrate/20170103085726_create_cashboxes.rb
+++ b/db/migrate/20170103085726_create_cashboxes.rb
@@ -1,0 +1,9 @@
+class CreateCashboxes < ActiveRecord::Migration
+  def change
+    create_table :cashboxes do |t|
+      t.integer :balance, default: 0
+      t.references :cashboxable, polymorphic: true, index: true
+      t.timestamps null: false
+    end
+  end
+end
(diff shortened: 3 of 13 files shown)


Maroo-b:
@b264 I added unit tests to the changes. 
I'm curious to know about the value of model association tests ? Do you often add association tests ?

NikitaAvvakumov:
@Maroo-b re: association tests - my 2 cents is that they have marginal utility during TDD but lose all value once code has been written

Maroo-b:
@NikitaAvvakumov thanks for the answer :) 
From my side I usually avoid adding tests for association because I have the feeling that I'm testing the framework, and in case I added them in strict TDD session I vote for removing them afterwards.

spec/factories/cashboxes.rb
```diff
@@ -0,0 +1,6 @@
+FactoryGirl.define do
+  factory :cashbox do
+    balance 1
```
b264: I'd suggest `balance { FactoryHelper::MySQL.integer }` here

app/models/cashbox.rb
```diff
@@ -0,0 +1,7 @@
+class Cashbox < ActiveRecord::Base
+  has_many :deposits, class_name: 'Payment', foreign_key: 'destination_id'
+  has_many :withdrawls, class_name: 'Payment', foreign_key: 'source_id'
```
NikitaAvvakumov: What's the advantage of having separate `deposits` & `withdrawals` associations? Why not just have `:payments` (actually, I would have preferred `:transactions` - perhaps we can still make this change?) whose value is either negative or positive? It doesn't seem right that to get all transactions/payments for a single cashbox, we need to query for both deposits & withdrawals separately. In this case, if there is still some utility in being able to retrieve just the deposits or just the withdrawals, we can create scopes that filter transactions based on whether their value is positive or negative.

db/migrate/20170103124956_add_fk_to_payments.rb
```diff
@@ -0,0 +1,9 @@
+class AddFkToPayments < ActiveRecord::Migration
+  def change
+    add_column :payments, :destination_id, :integer
+    add_column :payments, :source_id, :integer
+
+    add_index :payments, :destination_id
+    add_index :payments, :source_id
+  end
+end
```
NikitaAvvakumov: Need corresponding changes to the `Payment` model

spec/models/cashbox_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+
+RSpec.describe Cashbox, type: :model do
+
+  it { is_expected.to validate_numericality_of(:balance).only_integer }
+  it { is_expected.to have_many :deposits }
```
NikitaAvvakumov: As per our discussion in Slack, recommend getting rid of the association tests

db/migrate/20170103085726_create_cashboxes.rb
```diff
@@ -0,0 +1,9 @@
+class CreateCashboxes < ActiveRecord::Migration
+  def change
+    create_table :cashboxes do |t|
+      t.integer :balance, default: 0
```
NikitaAvvakumov: As per Slack, I think storing `balance` on `Transaction`/`Payment` rather than `Cashbox` is a more robust solution

app/models/cashbox.rb
b264: I agree, instead of Cashbox having two foreign ids maybe each Payment has two foreign ids-- source and destination

db/migrate/20170103085726_create_cashboxes.rb
b264: For the record, I'm still not convinced we should store balance at all.  It is calculable.  This seems like a premature scaling optimisation.  If everyone else agrees that we should store it in the db, then we need to ensure only the model itself can write to it, eg `def balance=; raise; end`  When and if the time comes where this is causing db thrashing then we can address it separately YAGNI

app/models/cashbox.rb
Maroo-b: @b264 I'm already using  two  two foreign ids-- source and destination on payment table via this [migration](https://github.com/AgileVentures/osra/pull/441/files#diff-022e8f544d24cf4c36d9116d2fd0c1c3) 

@NikitaAvvakumov I guess that payment should be a positive value between source and destination for example if we have a transaction/payment between two cashbox and we set a negative value (in this case this value represents a withdrawal from a cashbox which is negative value but it's in the same time a deposit to the destination cashbox which should be positive)
That's why I opted for two separate associations to keep positive value and use FK on payment model ( source and destination)

db/migrate/20170103124956_add_fk_to_payments.rb
Maroo-b: @NikitaAvvakumov sorry but  I didn't get the note, what changes should be done to `Payment` model ?

db/migrate/20170103085726_create_cashboxes.rb
Maroo-b: For the moment let's  have  amount on payment/transaction as we agreed the first time to move forward with the feature.

app/models/cashbox.rb
b264: @Maroo-b Yes, I see.  I like the way you did it, aside from disagreeing with storing a balance and the Payment model isn't updated with its `belongs_to` associations

db/migrate/20170103124956_add_fk_to_payments.rb
b264: calls to `ActiveRecord::Base.belongs_to` need to be added

Maroo-b:
I updated the PR to calculate the balance from payment, and I didn't add the validation on `Payment` destination FK because there is the case of withdrawal from sponsorship cashbox.

app/models/payment.rb
```diff
@@ -5,5 +5,8 @@
 # "amount" is in integer quantity in halalas, or hundredths of a riyal.
 
 class Payment < ActiveRecord::Base
+  belongs_to :source, class_name: 'Cashbox'
+  belongs_to :destination, class_name: 'Cashbox'
+
   validates :amount, numericality: { only_integer: true }, allow_nil: false
```
NikitaAvvakumov: With the deposits/withdrawals implementation, the amount should never be negative, correct? If so, we should add validation for that

Maroo-b:
@NikitaAvvakumov validation for `Payment` amount is added

db/migrate/20170103124956_add_fk_to_payments.rb
b264: :+1:

PR #443, opened 2017-01-09
Create method to make payments into sponsor's cashbox

Maroo-b:
Addresses: https://osraav.atlassian.net/browse/OSRA-453

### DIFF
--- a/app/services/payments/create_funds.rb
+++ b/app/services/payments/create_funds.rb
@@ -6,17 +6,19 @@ module Payments
     end
 
     def call
-      @payment.destination = @cashbox
-      @payment.save
+      if @payment.valid?
+        @payment.destination = @cashbox
+        @payment.save
+      end
     end
 
     private
 
     attr_reader :cashbox, :payment
 
-    def initialize(cashbox, payment)
+    def initialize(cashbox, amount, payment_builder = Payment)
       @cashbox = cashbox
-      @payment = payment
+      @payment = payment_builder.new(amount: amount)
     end
   end
 end
--- a/app/services/payments/create_funds.rb
+++ b/app/services/payments/create_funds.rb
@@ -1,0 +1,22 @@
+module Payments
+  class CreateFunds
+
+    def self.call(*args)
+      new(*args).call
+    end
+
+    def call
+      @payment.destination = @cashbox
+      @payment.save
+    end
+
+    private
+
+    attr_reader :cashbox, :payment
+
+    def initialize(cashbox, payment)
+      @cashbox = cashbox
+      @payment = payment
+    end
+  end
+end
--- a/app/services/payments/create_funds.rb
+++ b/app/services/payments/create_funds.rb
@@ -6,19 +6,21 @@ module Payments
     end
 
     def call
-      if @payment.valid?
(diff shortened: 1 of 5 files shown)


spec/factories/payments.rb
```diff
@@ -2,6 +2,6 @@
 # frozen_string_literal: true
 FactoryGirl.define do
   factory :payment do
-    amount { FactoryHelper::MySQL.integer }
+    amount { FactoryHelper::MySQL.integer(unsigned: true) }
```
b264: `unsigned int` goes to 4294967295 which will overflow.  Try `FactoryHelper::MySQL.int(min: 0, max: 2147483647)`

app/services/payments/create_funds.rb
```diff
@@ -0,0 +1,22 @@
...
+    def call
+      @payment.destination = @cashbox
+      @payment.save
+    end
+
+    private
+
+    attr_reader :cashbox, :payment
+
+    def initialize(cashbox, payment)
```
b264: I'd use kwargs here for code clarity `def initialize cashbox:, payment:`

spec/services/payments/create_funds_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+module Payments
+  describe CreateFunds do
+    describe '.call' do
+      context 'with positive amount' do
+        it 'add funds to a cashbox' do
+          cashbox = Cashbox.create
+
+          res = CreateFunds.call(cashbox, 1000)
+
```
b264: should we also be checking `expect(res).to be_truthy`, or is that not an intended behaviour?

app/services/payments/create_funds.rb
```diff
@@ -0,0 +1,26 @@
+module Payments
+  class CreateFunds
+
+    def self.call(*args)
+      new(*args).call
+    end
+
+    def call
+      payment = payment_builder.create(amount: amount)
```
b264: using create here instead of new creates one extra db write, but the bigger problem is that all the db writes then need to be in an atomic transaction

Maroo-b:
@b264 I reverted to the old implementation by instantiating a `Payment` instance instead of creating it

spec/models/cashbox_spec.rb
```diff
@@ -3,18 +3,28 @@
...
+      it 'returns balance of cashbox' do
+        cashbox_source = Cashbox.create
+        cashbox_destination = Cashbox.create
 
-      deposit_without_source = create(:payment, destination: cashbox_destination,
-                                      amount: 150)
-      deposit_with_source = create(:payment, destination: cashbox_destination,
-                                   source: cashbox_source, amount: 100)
-      withdrawal = create(:payment, amount: 50, source: cashbox_destination)
+        deposit_without_source = create(:payment, destination: cashbox_destination,
```
b264: Is there any way to use something like
```
let(:amount_1) { FactoryHelper::MySQL.int(min: 1, max: 2147483647) }
```
for these three amounts instead of hardocded integers?

b264:
Please reopen this PR into the `450_track_payments` branch instead of `develop`

PR #444, opened 2017-01-11
Fix security warning from Gemnasium

Maroo-b:
- Addresses: https://osraav.atlassian.net/browse/OSRA-458

@b264 can you launch a new diagnostic from [Gemnasium](https://gemnasium.com/github.com/b264/osra) ?


Maroo-b:
Replace PR with https://github.com/AgileVentures/osra/pull/446

PR #446, opened 2017-01-15
Update dependencies

NikitaAvvakumov:
- update all gems while leaving Ruby at 2.2.0 & Rails at 4.2.7
- update code as needed
- minor refactoring

1. Leaving Gemfile without any versions for now because intend to immediately upgrade Rails to 5.x & Ruby to 2.3/2.4. Then will add version locks.
2. roo gem (spreadsheet parsing) raises deprecation warnings because of the way the roo-xls gem calls its code. roo-xls is no longer maintained. We can live with the warnings for now, but at some point these will turn into exceptions. Can either fork & fix roo-xls (there is already a PR with the fix in the repo) or perhaps client doesn't need to work with the old XLS format any longer.
3. To test, use https://osra-nga.herokuapp.com/, admin@example.com, Password1. Note that all old orphan lists have been deleted from S3. The error you get when attempting to download them is caused by this, not new gem versions. For testing, "partner100" has one list uploaded today that you can download, or upload your own.


spec/rails_helper.rb
```diff
@@ -52,3 +52,10 @@
 
   config.global_fixtures = :all
 end
+
+Shoulda::Matchers.configure do |config|
+  config.integrate do |with|
+    with.test_framework :rspec
+    with.library :rails
+  end
+end
```
Maroo-b: @NikitaAvvakumov what do you think about adding the config to **spec/support/shoulda_matchers.rb**  ? I saw in many Rails projects that developers tend to put RSpec additional config in **spec/support** folder

NikitaAvvakumov: I have no strong feelings one way or the other, at this point. I think it's a good idea to keep an eye on this and consider taking action if the helper file starts getting out of hand. At the moment, though, it's 60 lines of mostly comments, so I'm OK to let it be

NikitaAvvakumov:
@Maroo-b All specs are currently passing. Going forward, I would target controller specs for (selective) removal first. All specs could use a refactoring to decouple them from the implementation, avoid testing the framework, and generally test meaningful things only. It's probably best to do this separately from any other task or feature, not least so that we don't develop a bad habit of deleting failing specs instead of trying to fix them 😆

Maroo-b:
That's interesting, usually I don't like fixing view spec and I opt for deleting it, maybe I should reconsider this habit :thinking:

